PR #5, opened 2018-03-15
add aiida-vasp to python requirements

DropD:


ltalirz:
@DropD On second thought... there is a '=' missing... can you please add this?

PR #22, opened 2019-09-18
Aiida 1.0 k8s

yakutovicha:


yakutovicha:
@ltalirz so I tested on dev-aiidalab and it works. 

Feel free to squash and merge the PR. If you want to test it before - go to your dev-aiidalab account and restart the server.

ltalirz:
> @ltalirz so I tested on dev-aiidalab and it works.

Fantastic, thanks a lot!!  

I'll try to give it a try on kubernetes tonight. 
If I don't manage until tomorrow, feel free to merge.

ltalirz:
@yakutovicha your last "fix" broke travis
https://travis-ci.org/aiidalab/aiidalab-docker-stack/builds/586708570?utm_source=github_status&utm_medium=notification

I see that it is a change of the UID - this is certainly an issue for kubernetes, where the UID needs to be 1000 - perhaps this is also the problem with travis

ltalirz:
for the aiidalab on openstack you can "fix" this simply by passing the desired UID as a build argument when you build the docker container.
for kubernetes, unfortunately we will need a pre-built image with the correct UID

ltalirz:
see here: https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg

yakutovicha:
> for the aiidalab on openstack you can "fix" this simply by passing the desired UID as a build argument when you build the docker container.
> for kubernetes, unfortunately we will need a pre-built image with the correct UID

alright, so then probably just delete the last commit. Otherwise, everything works fine. The UID will be provided in the ansible scripts then.

ltalirz:
> The UID will be provided in the ansible scripts then.

You could even include it in `build.sh` since this is anyhow not used by Dockerhub.

ltalirz:
I just gave your image a try (it anyhow built only the previous commit, since this one did not pass travis).

Unfortunately, the problem remains (was also to be expected):
```
+ export SHELL=/bin/bash
+ SHELL=/bin/bash
+ reentry scan
+ aiida_backend=django
+ '[' '!' -d /project/.aiida ']'
+ verdi setup --profile default --non-interactive --email some.body@xyz.com --first-name Some --last-name Body --institution XYZ --db-backend django --db-username aiida --db-password aiida_db_passwd --db-name aiidadb --db-host localhost --db-port 5432 --repository /project/aiida_repository
Success: created new profile `default`.
Info: migrating the database.
Critical: database migration failed, probably because connection details are incorrect:
could not connect to server: Connection refused
	Is the server running on host "localhost" (127.0.0.1) and accepting
	TCP/IP connections on port 5432?
could not connect to server: Cannot assign requested address
	Is the server running on host "localhost" (::1) and accepting
	TCP/IP connections on port 5432?
```

ltalirz:
It must be something related to the way postgresql is running (or not running) that was changed between `develop` and here.

ltalirz:
the strange thing is that you tested the new contained on the aiidalab and it seems to be running fine, right?
could it be related to the time the postgres server needs to start?
that in one case it is already started when it arrives to `verdi setup`, while in another case it hasn't yet started?

ltalirz:
By the way - what was again the reason that postgresql is not started as a service like rabbitmq and the daemon?

ltalirz:
Ok, now the image with UID 1234 has been pushed to dockerhub as well - this should be changed.

There are a few occurences of `scientist` as well that should be changed to the variable (but this is not causing the issue as we're currently not modifying the variable)
```
$ git grep -i scientist
.travis.yml:    - container_user: scientist
Dockerfile:ARG NB_USER="scientist"
README.md:su scientist
inspect.sh:# login as scientist
inspect.sh:docker run --init --user scientist -ti aiidalab-docker-stack:develop /bin/bash
my_init.d/start-postgres.sh:su -c /opt/postgres.sh scientist
my_init.d/start-singleuser.sh:su -c /opt/start-singleuser.sh scientist
service/aiida:su -c 'verdi daemon start --foreground' scientist
service/jupyterhub-singleuser:su -c  /opt/start-jupytehub-singleuser.sh scientist
```

yakutovicha:
> By the way - what was again the reason that postgresql is not started as a service like rabbitmq and the daemon?

To avoid exactly the problem you have: postgrest must be ready before the user's configuration part. Services start after it.

yakutovicha:
> the strange thing is that you tested the new container on the aiidalab and it seems to be running fine, right?
> could it be related to the time the postgres server needs to start?
> that in one case it is already started when it arrives to `verdi setup`, while in another case it hasn't yet started?

That is probably the reason, I think. I don't know what exactly happens in Kubernetes, but in my current setup, I enforce Postgres to start before the user's configuration part.

yakutovicha:
> There are a few occurences of `scientist` as well that should be changed to the variable (but this is not causing the issue as we're currently not modifying the variable)

I wonder what is the right approach to achieve this. If I am going to base this container on top of the `aiida-docker-stack`, which is already pre-built, I can't really modify such a variable there.




> > The UID will be provided in the ansible scripts then.
> 
> You could even include it in `build.sh` since this is anyhow not used by Dockerhub.

I don't use it in Ansible role either.

yakutovicha:
> > The UID will be provided in the ansible scripts then.
> 
> You could even include it in `build.sh` since this is anyhow not used by Dockerhub.

I don't use it in Ansible role either.

PR #64, opened 2020-04-21
Do not install jupyter in conda environment.

yakutovicha:
Previously, everything was installed in the base conda environment,
however, this created some problems in coupling jupyter with AiiDA
which requires tornado to be lower than 5.0.0. To fix this problem
this PR installs jupyter in a 'root' environment, but configures it
to use the ipykernel from the base conda environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,8 @@ RUN /usr/bin/pip3 install          \
     'nglview'                      \
     'voila'
 
+RUN python -m ipykernel install
+
 # Enable extensions.
 RUN /usr/local/bin/jupyter serverextension enable --py --sys-prefix nbserverproxy
 RUN /usr/local/bin/jupyter nbextension     enable --py --sys-prefix appmode
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,13 +51,13 @@ RUN /usr/bin/pip3 install          \
     'voila'
 
 # Enable extensions.
-RUN /usr/local/bin/jupyter serverextension enable --sys-prefix --py nbserverproxy
+RUN /usr/local/bin/jupyter serverextension enable --py --sys-prefix nbserverproxy
 RUN /usr/local/bin/jupyter nbextension     enable --py --sys-prefix appmode
 RUN /usr/local/bin/jupyter serverextension enable --py --sys-prefix appmode
 RUN /usr/local/bin/jupyter nbextension enable nglview --py --sys-prefix
 # TODO: delete, when https://github.com/aiidalab/aiidalab-widgets-base/issues/31 is fixed
 # the fileupload extension also needs to be "installed".
-RUN /usr/local/bin/jupyter nbextension install --sys-prefix --py fileupload
+RUN /usr/local/bin/jupyter nbextension install --py --sys-prefix fileupload
 
 # Enables better integration with Jupyter Hub.
 # https://jupyterlab.readthedocs.io/en/stable/user/jupyterhub.html#further-integration
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,8 +50,6 @@ RUN /usr/bin/pip3 install          \
     'nglview'                      \
     'voila'
 
-RUN python -m ipykernel install --user
-
 # Enable extensions.
 RUN /usr/local/bin/jupyter serverextension enable --sys-prefix --py nbserverproxy
 RUN /usr/local/bin/jupyter nbextension     enable --py --sys-prefix appmode
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,9 @@ RUN apt-get update && apt-get install -y  \
     file                  \
     libssl-dev            \
     libffi-dev            \
+    python3-pip           \
+    python3-setuptools    \
+    python3-wheel         \
     quantum-espresso      \
   && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -34,24 +37,33 @@ chown -R root:root /opt/pseudos/;                                  \
...
-RUN jupyter serverextension enable --sys-prefix --py nbserverproxy
+    'fileupload==0.1.5'            \
+    'nbserverproxy==0.8.8'         \
+    'appmode-aiidalab==0.5.0.1'    \
+    'notebook==5.7.8'              \
+    'nglview'                      \
+    'voila'
+
+# Enable extensions.
+RUN /usr/local/bin/jupyter serverextension enable --sys-prefix --py nbserverproxy
```
ltalirz, line 54: Just to improve readability
```suggestion
RUN /usr/local/bin/jupyter serverextension enable --py --sys-prefix nbserverproxy
```

Dockerfile
```diff
@@ -34,24 +37,33 @@ chown -R root:root /opt/pseudos/;                                  \
...
+RUN /usr/local/bin/jupyter nbextension     enable --py --sys-prefix appmode
+RUN /usr/local/bin/jupyter serverextension enable --py --sys-prefix appmode
+RUN /usr/local/bin/jupyter nbextension enable nglview --py --sys-prefix
+# TODO: delete, when https://github.com/aiidalab/aiidalab-widgets-base/issues/31 is fixed
+# the fileupload extension also needs to be "installed".
+RUN /usr/local/bin/jupyter nbextension install --sys-prefix --py fileupload
```
ltalirz, line 60:
```suggestion
RUN /usr/local/bin/jupyter nbextension install --py --sys-prefix fileupload
```

Dockerfile
```diff
@@ -34,24 +37,33 @@ chown -R root:root /opt/pseudos/;                                  \
...
     'jupyterlab==0.35.4'           \
-    'nbserverproxy==0.8.8'
-
-# Enable nbserverproxy extension.
-RUN jupyter serverextension enable --sys-prefix --py nbserverproxy
+    'fileupload==0.1.5'            \
+    'nbserverproxy==0.8.8'         \
+    'appmode-aiidalab==0.5.0.1'    \
+    'notebook==5.7.8'              \
+    'nglview'                      \
```
ltalirz, line 50: just out of curiosity: why is this needed here?

Dockerfile
```diff
@@ -34,24 +37,33 @@ chown -R root:root /opt/pseudos/;                                  \
...
+    'voila'
+
+# Enable extensions.
+RUN /usr/local/bin/jupyter serverextension enable --sys-prefix --py nbserverproxy
+RUN /usr/local/bin/jupyter nbextension     enable --py --sys-prefix appmode
```
ltalirz, line 55: also here: why is this needed now, but wasn't needed before?

Dockerfile
```diff
@@ -34,24 +37,33 @@ chown -R root:root /opt/pseudos/;                                  \
...
 # Enables better integration with Jupyter Hub.
 # https://jupyterlab.readthedocs.io/en/stable/user/jupyterhub.html#further-integration
 # Takes about 3 minutes and 20 seconds.
 RUN jupyter labextension install @jupyterlab/hub-extension
 
-# TODO: delete, when https://github.com/aiidalab/aiidalab-widgets-base/issues/31 is fixed
-# the fileupload extension also needs to be "installed".
-RUN jupyter nbextension install --sys-prefix --py fileupload
-
 # Install jupyterlab theme.
```
ltalirz, line 67: by the way, as long as we aren't using it, I think it's perfectly fine to comment out anything related to the jupyterlab
(not sure whether the latest voilà stuff is starting to use it?)

opt/prepare-aiidalab.sh
```diff
@@ -91,3 +91,6 @@ if [ ! -e /home/${SYSTEM_USER}/apps ]; then
   git checkout ${AIIDALAB_DEFAULT_GIT_BRANCH}
  cd -
 fi
+
+# Activate the right python kernel.
+python -m ipykernel install --user
```
ltalirz, line 96: can this go in the dockerfile with the other python-setup-related commands?

Dockerfile
yakutovicha: this was a workaround to be able to run https://github.com/aiidalab/aiidalab-docker-stack/blob/821e78c00677e5f6139375da1c633cc643524657/Dockerfile#L57
which, in turn, was necessary to have `nglview` working.

yakutovicha: To be clear: I am aware that this is a hack. Not sure, though, what is the better way to achieve the same. One needs to play around with it.

Dockerfile
yakutovicha: Well, without this appmode won't work. Not sure why, though.

The docker container that Ole had created actually contained that line. His container was based on the same principles we are now basing the new container on - keeping Jupyter stuff in a separate python environment.

When I was moving stuff to AiiDA 1.0, I removed it because I moved everything in the same Python3 environment. Now I am moving things back just to give another evidence that _history repeats itself_ ;)

opt/prepare-aiidalab.sh
yakutovicha: It was originally in the Docker file, but then I moved it here. The reason was this `--user` option. But it is not clear whether the option is at all needed. I will now try it out without `--user`.

Dockerfile
yakutovicha: I think Dou might start using it, so let's keep it.

yakutovicha:
> Could you please also let me know the easiest instructions for testing it locally?
> Can I just follow the readme?

No, I think we should add an option to build things locally, in the current implementation `./run_aiidalab.sh` pulls the image from the Docker Hub. 

you should do:
```bash
./build.sh
docker run -d -p 9999:8888 -e JUPYTER_TOKEN="PUTH_YOUR_TOKEN_HERE" -v /Users/yakutovicha/aiidalab_test:/home/aiida aiidalab-docker-stack:develop
```

opt/prepare-aiidalab.sh
yakutovicha: worked, so I moved it back to the docker file in ac4d402

Dockerfile
yakutovicha: done

PR #66, opened 2020-04-24
Use openssl for token generation.

csadorf:
Inspired by: https://jupyterhub.readthedocs.io/en/stable/getting-started/security-basics.html#generating-and-storing-as-a-cookie-secret-file

Tested for for Ubuntu 18.04 and on Mac OS-X 10.11.6.

Caveat: Requires the installation of openssl at least on Mac OS-X.

### DIFF
--- a/run_aiidalab.sh
+++ b/run_aiidalab.sh
@@ -12,7 +12,7 @@ fi
 
 PORT=${1}
 FOLDER=${2}
-TOKEN=`date | md5`
+TOKEN=`openssl rand -hex 32`
 IMAGE='aiidalab/aiidalab-docker-stack:latest'
 
 echo "Pulling the image from the Docker Hub..."


yakutovicha:
hi @csadorf, have you seen #65? ;-D

csadorf:
> hi @csadorf, have you seen #65? ;-D

I did not, however it seems to me that #65 introduces additional changes?

yakutovicha:
> > hi @csadorf, have you seen #65? ;-D
> 
> I did not, however it seems to me that #65 introduces additional changes?

The OpenSSL idea is similar, but I agree that your approach is better, so I am going to merge it.

 
Yes, for some reason on ubuntu machine of my colleague ':latest' suffix caused the script to fail. This is why I deleted that.

PR #76, opened 2020-06-02
Bump several package versions:

yakutovicha:
jupyterhub from 0.9.4 to 1.1.0
jupyterlab from 0.35.4 to 2.1.4
notebook from 5.7.8 to 6.0.3

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,11 +65,6 @@ RUN /usr/local/bin/jupyter nbextension enable nglview --py --sys-prefix
 # the fileupload extension also needs to be "installed".
 RUN /usr/local/bin/jupyter nbextension install --py --sys-prefix fileupload
 
-# Enables better integration with Jupyter Hub.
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -43,12 +43,12 @@ RUN pip install 'aiidalab==v20.05.0b1'
...
+    'jupyterhub==1.1.0'            \
+    'jupyterlab==2.1.4'            \
     'fileupload==0.1.5'            \
     'nbserverproxy==0.8.8'         \
-    'appmode-aiidalab==0.5.0.1'    \
-    'notebook==5.7.8'              \
+    'appmode==0.7.0'               \
+    'notebook==6.0.3'              \
     'nglview'                      \
     'voila'
```
ltalirz, line 53: This seems unsafe. Can you fix a version here?
Same for nglview - since you mentioned this is somehow a hack and needs to be in sync with the nglview used inside the conda env, it is even more important to lock down the version

PR #80, opened 2020-06-08
Appmode

### LINKED ISSUE #67: Closed apps' kernel keeps running
The right behaviour should be stopping the kernel when closing the corresponding tab. Currently, this does not happen.

yakutovicha:
fixes #67 

1) Install appmode directly from git
2) Replace the gears by the ones of AiiDA lab


Dockerfile
```diff
@@ -93,6 +90,13 @@ COPY my_init.d/prepare-aiidalab.sh /etc/my_init.d/80_prepare-aiidalab.sh
 COPY opt/start-notebook.sh /opt/
 COPY service/jupyter-notebook /etc/service/jupyter-notebook/run
 
+# Activate appmode
+WORKDIR /opt/
+RUN git clone https://github.com/oschuett/appmode.git
```
ltalirz, line 95: Please check out a fixed version

yakutovicha: We need to wait until the version is released by Ole.

ltalirz: you can just use the sha1 hash for the moment

yakutovicha:
> it means we'll need to do the same for the Quantum Mobile as well... and remove aiidalab-appmode from the aiidalab meta package

That is already done. As soon as jupyter and aiida environments got split, appmode wasn't needed any longer in the aiidalab package.

ltalirz:
> That is already done. As soon as jupyter and aiida environments got split, appmode wasn't needed any longer in the aiidalab package.

yakutovicha:
> Ok, great this fixes it for the aiidalab meta package. I guess this still means we need to fix it on quantum mobile, though (where we install it inside the VM)

yes

PR #85, opened 2020-06-19
Enable continuous integrated testing

csadorf:
Enable continuous integrated testing for this repository.

The tests are implemented as part of the `continuous-integration` GitHub actions workflow which consists of two jobs:

1. Build the docker image and push it to docker hub tagged with both the ref (branch name) and the sha1.
2. Run the `aiidalab-test-app-action` action for a hard-coded set of continuously tested apps.

The list of apps should be extended in the future and we might need to devise a different mechanism of generating that list, but for now this will suffice.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,36 +7,36 @@ on:
 
 jobs:
 
-  build-docker-image:
+  #build-docker-image:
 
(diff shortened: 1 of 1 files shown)


csadorf:
Replaced by https://github.com/aiidalab/aiidalab-docker-stack/pull/86 .

PR #86, opened 2020-06-19
Enable continuous integrated testing with GitHub actions workflow.

csadorf:
Enable continuous integrated testing for this repository.

The tests are implemented as part of the `continuous-integration` GitHub actions workflow which consists of two jobs:

1. Build the docker image and push it to docker hub tagged with both the ref (branch name) and the sha1.
2. Run the [aiidalab-test-app-action](https://github.com/aiidalab/aiidalab-test-app-action) for a hard-coded set of continuously tested apps.

The list of apps should be extended in the future and we might need to devise a different mechanism of generating that list, but for now this will suffice.

The home app is currently failing which is why the tests for the home app are disabled.
However I'd like to merge this anyways and then fix the issues with the home app as part of a separate PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,65 @@
+# Run basic tests for this app on the latest aiidalab-docker image.
+
+name: continuous-integration
+
+on:
+  [push]
+
+jobs:
+
+  build-docker-image:
+
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    outputs:
+      image: ${{ steps.tag.outputs.image }}
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: build image
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          tag_with_ref: true
+          tag_with_sha: true
+          cache_froms: "aiidalab/aiidalab-docker-stack:latest,aiidalab/aiidalab-docker-stack:stable,aiidateam/aiida-core:latest"
+
+      - name: set output image
+        id: tag
+        run: |
+          IMAGE_REF=`echo ${{ github.sha }} | sed 's/^\(.......\).*/\1/'`
+          echo "::set-output name=image::${{ github.repository }}:sha-${IMAGE_REF}"
+
+  test-apps:
+
+    runs-on: ubuntu-latest
+    needs: build-docker-image
+    timeout-minutes: 10
+
+    strategy:
+      matrix:
+        app: [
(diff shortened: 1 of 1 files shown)


csadorf:
Not sure what the issue with travis is. The test succeeded on the previous commit.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
+# Run basic tests for this app on the latest aiidalab-docker image.
```
ltalirz, line 1: Delete?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+      - uses: actions/checkout@v2
+
+      - name: build image
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          tag_with_ref: true
+          tag_with_sha: true
+          cache_froms: "aiidalab/aiidalab-docker-stack:latest,aiidalab/aiidalab-docker-stack:stable,aiidateam/aiida-core:latest"
```
ltalirz, line 28: ah, didn't know about this. very nice!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+
+    steps:
+
+      - name: Checkout app
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ matrix.app }}
+
+      - name: Test app
+        uses: aiidalab/aiidalab-test-app-action@v2
```
ltalirz, line 62: would you mind adding one line of comment here on what this actually tests?

.github/workflows/ci.yml
csadorf: Unfortunately it does not work as well as expected. I've investigated ways to improve the performance of docker builds in the context of GitHub actions workflows and my conclusion was that any attempt to introduce "manual caching" is likely not going to lead to any significant performance improvement, but is going to massively increase complexity. The best chance is that GitHub implement layered caching as part of their execution platform, which is likely going to happen at some point, until then we will likely need to live with the current build times.

.github/workflows/ci.yml
csadorf: I've generally overhauled the documentation of the workflow definition.

csadorf:
@ltalirz Thank you for your comments. I've generally overhauled the documentation of the workflow definition in response. I plan on gradually extending the documentation of our testing framework in the upcoming weeks. Specifically, in the context of implementing more unit tests for our backend (platform and apps), and it will probably also worthwhile to implement a few more "complex" front-end tests. Finally, we might want to consider running cron job tests against our staging and production environments. However, all of this is out-of-scope for this PR.

PR #91, opened 2020-07-16
Enable the custom widgets with Voila

dou-du:
This will allow to show custom widgets with Voila, for instance the widget periodic table.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -89,7 +89,8 @@ RUN /usr/bin/pip3 install voila==0.1.21
 RUN /usr/bin/pip3 install voila-aiidalab-template==0.0.2
 
 # Enable widget_periodictable (installed with aiidalab package).
-RUN jupyter nbextension enable --py --system widget_periodictable
+RUN /usr/local/bin/jupyter nbextension install --py --user widget_periodictable
+RUN /usr/local/bin/jupyter nbextension enable widget_periodictable --user --py
 
 # Install some useful packages that are not available on PyPi
 RUN conda install --yes -c conda-forge rdkit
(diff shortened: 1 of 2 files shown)


dou-du:
> Thanks, @dou-du, I have one question and one suggestion.
> 
> q: did you try to run it yourself? The periodic table stopped working for me, even with the app mode.
> s: Can you please add it for the JupyterHub part as well?

I cannot run it either. I need to run two additional command lines:

```bash
jupyter nbextension install --py --user widget_periodictable
jupyter nbextension enable widget_periodictable --user --py
```
I just update it in the Dockerfile and make the pull request here. I am not sure about whether it will fix the problem. Could you recompile the docker image and test it ? Thanks.

yakutovicha:
I tested it locally, but it didn't work. Please try to fix the problems. Here are the commands to test things locally

```shell
PORT=8888  # Local port
IMAGE='aiidalab-docker-stack:develop'  # image name
FOLDER="${HOME}/aiidalab"  # Folder to be mounted
docker build -t ${IMAGE} ./  # To build the container
CONTAINERID=`docker run -d -p ${PORT}:8888 -e JUPYTER_TOKEN=${TOKEN} -v "${FOLDER}":/home/aiida ${IMAGE}`  #To start the container
docker exec --tty ${CONTAINERID} wait-for-services  # To wait until the container is started
```

yakutovicha:
@dou-du, it should be working now. Do you want to have a look?

dou-du:
> @dou-du, it should be working now. Do you want to have a look?

Thanks a lot. I will check it later today.

PR #98, opened 2020-08-06
Install OPTIMADE client

yakutovicha:


yakutovicha:
I tested it locally, seems to work fine!

Dockerfile
```diff
@@ -93,6 +93,11 @@ RUN /usr/bin/pip3 install widget-periodictable==2.1.2
 RUN /usr/local/bin/jupyter nbextension install --py --user widget_periodictable
 RUN /usr/local/bin/jupyter nbextension enable widget_periodictable --user --py
 
+# Install OPTIMADE.
+WORKDIR /opt/
+RUN git clone https://github.com/aiidalab/aiidalab-optimade.git && cd aiidalab-optimade && git reset --hard e008ca3f00bfbcceea52512e7dfe1c24f803c775
```
CasperWA, line 98: Did you fix it to a commit SHA as a way to fix to a version or?
```suggestion
RUN git clone https://github.com/aiidalab/aiidalab-optimade.git && cd aiidalab-optimade && git checkout origin/aiidalab
```

Dockerfile
```diff
@@ -93,6 +93,11 @@ RUN /usr/bin/pip3 install widget-periodictable==2.1.2
...
+RUN pip install ./aiidalab-optimade
```
CasperWA, line 99: Even though you're entering the `aiidalab-optimade` folder in the previous `RUN`, this is disregarded due to the `WORKDIR` command, or?

Dockerfile
yakutovicha: Yes, this is a way to fix the version

Dockerfile
yakutovicha: correct, once the command is executed you return back to the folder specified in the previous `WORKDIR` call.

Dockerfile
CasperWA: Would it be a problem or issue if you just took the aiidalab branch only? This way it would get all the updates I will do as well...

CasperWA: Or do you prefer to specifically use a commit SHA here?

yakutovicha: Typically, I update everything all together. This way we can link the specific version of AiiDA lab image with the installed packages. That is why SHA is preferable.

Of course, more packages we have, more hassle it becomes. This is why it is important to enable the installation of the app's dependencies.

CasperWA: Yeah, that was my point (the latter). But if you want to keep the commit SHA that's fine then. Just be aware that it's work in progress and might change quickly over very few days...

yakutovicha: Sure, before releasing a new image I will go through the packages and update their SHA versions.

And, if I may add, releasing OPTIMADE widget as a pypi package would be super-extra great. This way I can add it as a proper dependency, which would allow for a proper dependency resolution.

PR #102, opened 2020-08-18
Bump aiida-core base image to 1.3.0.1

yakutovicha:


yakutovicha:
> What is the motivation?

Incorporate the following bug fix: https://github.com/aiidateam/aiida-core/issues/4289

PR #108, opened 2020-08-26
Don't install dscribe, update OPTIMADE.

yakutovicha:
Dscribe package pulls ASE==3.19.1 as the dependency, which overrides
ASE==3.20.1 installed as the dependency of the aiidalab package. In
the future the users who want to use dscribe may install it themselves.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -95,7 +95,7 @@ RUN /usr/local/bin/jupyter nbextension enable widget_periodictable --user --py
 
 # Install OPTIMADE.
 WORKDIR /opt/
-RUN git clone https://github.com/aiidalab/aiidalab-optimade.git && cd aiidalab-optimade && git reset --hard e008ca3f00bfbcceea52512e7dfe1c24f803c775
+RUN git clone https://github.com/aiidalab/aiidalab-optimade.git && cd aiidalab-optimade && git reset --hard 25d56cc8ae992044265aaaad4968d904b4afb487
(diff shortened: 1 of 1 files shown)


yakutovicha:
@CasperWA do you know why we can't install optimade anymore?

CasperWA:
> @CasperWA do you know why we can't install optimade anymore?

Yeah. This is due to my system of handling the repo. I see now it may be the wrong way that I'm doing it, since I didn't think that people relied on commit SHA's.
I usually rebase the `aiidalab` and `mcloud` branches on top of the latest `develop`. This will remove the commit SHA's you're pointing to. This is also why I originally wanted you to simply checkout/install the `aiidalab` branch. But maybe I should simply change my workflow here and instead merge new changes from `develop` into the two branches without rebasing...

yakutovicha:
> But maybe I should simply change my workflow here and instead merge new changes from `develop` into the two branches without rebasing...

yep, that would be ideal, cause in the current approach you are chaning the git history.

CasperWA:
> > But maybe I should simply change my workflow here and instead merge new changes from `develop` into the two branches without rebasing...
> 
> yep, that would be ideal, cause in the current approach you are chaning the git history.

That can be discussed, but it makes it much cleaner in the end for me.

CasperWA (approved):
I would still vote for installing the `aiidalab` branch HEAD commit instead of resetting to a specific commit SHA.

PR #111, opened 2020-08-31
Test apps with correct name and do not reinstall if bundled.

csadorf:
This enables the testing of pre-installed (bundled) apps directly at the
existing end point.


yakutovicha (approved):
looks good, thanks @csadorf

PR #118, opened 2020-09-14
Revise the run CLI

csadorf:
I've used the [argbash](https://argbash.io/) tool to create a new `run` interface. Besides having a slightly extended and more robust CLI, this script also changes the behavior to build the image and run said built image. The prior behavior of pulling the image from the hub is somewhat unexpected in my opinion, because it completely ignores the Dockerfile in this repository.

 - Rename run_aiidalab.sh -> run.sh .
 - Remove the build.sh script and roll function into run.sh.
 - Revise README to reflect changes.


yakutovicha (approved):
thanks @csadorf works very well. I just removed the `activate.sh` script as it is not needed anymore and rebased on top of the `develop` branch.

PR #121, opened 2020-09-17
Use v1.0.0 of aiidalab-optimade

CasperWA:
This is the new edition of the AiiDAlab-specific OPTIMADE Client application.

Note, depending on the outcome of the `numpy` dependency discussion, this version may change, hence this is a draft PR until that has been decided.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -95,7 +95,7 @@ RUN /usr/local/bin/jupyter nbextension enable widget_periodictable --user --py
 
 # Install OPTIMADE.
 WORKDIR /opt/
-RUN git clone https://github.com/aiidalab/aiidalab-optimade.git && cd aiidalab-optimade && git reset --hard v3.3.2+aiidalab
+RUN git clone https://github.com/aiidalab/aiidalab-optimade.git && cd aiidalab-optimade && git reset --hard v1.0.0
 RUN pip install ./aiidalab-optimade
 
 # Install some useful packages that are not available on PyPi


csadorf (commented):
Would it be possible to implement a simple test for the app? See one of the other apps for examples on how to do that.

yakutovicha (commented):
I am actually surprised that we need this at all. Now, since the OPTIMADe client is available through PyPi, I added it to the `requirements.txt` file of the aiidalab package. The OPTIMADE app can be installed through the App Store.

CasperWA:
> I am actually surprised that we need this at all. Now, since the OPTIMADe client is available through PyPi, I added it to the `requirements.txt` file of the aiidalab package. The OPTIMADE app can be installed through the App Store.

True, the main issue lies in the special `OptimadeQueryWidget`, which does not exist in the `optimade-client` package. But I guess I could implement it there and then there's no need for this?

yakutovicha:
> True, the main issue lies in the special `OptimadeQueryWidget`, which does not exist in the `optimade-client` package.

can it be the reason for the failing tests?

> But I guess I could implement it there and then there's no need for this?

exactly.

CasperWA:
> > True, the main issue lies in the special `OptimadeQueryWidget`, which does not exist in the `optimade-client` package.
> 
> can it be the reason for the failing tests?

In theory it shouldn't be, because the QE App import `OptimadeQueryWidget` from `aiidalab_optimade`, which would exactly be available through installing `aiidalab-optimade`.
However, with the change/removal of installing `aiidalab-optimade` here, the QE App should be updated to import from `optimade_client` instead, so at that point it _should_ fail.

> > But I guess I could implement it there and then there's no need for this?
> 
> exactly.

Opened CasperWA/voila-optimade-client#158 to deal with this.

csadorf:
Please ping me here or create an issue on the aiidalab-qe repositroy if there is any action I need to take on resolving this from the qe-app side.

CasperWA:
@yakutovicha So I have just come to the realization why it was separated.
It's due to the expectation of the `StructureManagerWidget` that the `structure` attribute traitlet of an importer is either an instance of `ase.Atoms` or `aiida_core.orm.Data`, while all the widgets in `optimade-client` uses its own instance of an OPTIMADE structure from `optimade.adapters.Structure`.

So the widget in `aiidalab-optimade` is indeed specific to AiiDAlab and how it's used there.
I could still create the widget and have it be importable from `optimade-client`, but one would still have to change the traitlet type and such, or do some `lambda` conversion in AiiDAlab always when using it?

CasperWA:
This is no longer valid, as it has been moved to `optimade-client` as well as `aiidalab-widgets-base`.

PR #134, opened 2020-11-02
Add Pipenv file to lock the python environment.

yakutovicha:
The dependencies are collected in the Pipenv file and further locked
using the `pipenv` tool. The locked dependencies are placed in the
requirements.txt file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -272,3 +272,9 @@ tags
 
 # End of https://www.gitignore.io/api/vim,linux,macos,python,pycharm
 submit_test
+
+# Custom.
+Pipfile.lock
+requirements.txt_dev
+requirements.txt_stable
+
--- a/.gitignore
+++ b/.gitignore
@@ -275,6 +275,3 @@ submit_test
 
 # Custom.
 Pipfile.lock
-requirements.txt_dev
-requirements.txt_stable
-
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,9 +55,12 @@ chmod -R +r /opt/pseudos/
 #     /usr/local/bin/jupyter labextension install *.tgz && \
 #    cd ..
 
+# Change workdir.
+WORKDIR /opt/
+
 # Install Python packages needed for AiiDAlab and populate reentry cache for root (https://pypi.python.org/pypi/reentry/).
-RUN pip install 'aiidalab==v20.10.0b1'
-#RUN pip install https://github.com/aiidalab/aiidalab/archive/243dacf18fc21e1cdfe89be2c3fe92b95d3172ef.zip
+COPY requirements.txt .
+RUN pip install -r requirements.txt
 RUN reentry scan
 
 # Install python kernel from the conda environment (comes with the aiidalab package).
@@ -72,7 +75,6 @@ RUN /usr/bin/pip3 install nglview==2.7.7
 RUN /usr/local/bin/jupyter nbextension enable nglview --py --sys-prefix
 
 # Install and enable appmode.
-WORKDIR /opt/
 RUN git clone https://github.com/oschuett/appmode.git && cd appmode && git reset --hard v0.8.0
 COPY gears.svg ./appmode/appmode/static/gears.svg
 RUN /usr/bin/pip3 install ./appmode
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -271,4 +271,10 @@ tags
...
 
 # End of https://www.gitignore.io/api/vim,linux,macos,python,pycharm
-submit_test
\ No newline at end of file
+submit_test
+
+# Custom.
+Pipfile.lock
+requirements.txt_dev
+requirements.txt_stable
```
csadorf, line 279: I don't think we need those (see other comment).

Dockerfile
```diff
@@ -72,7 +75,6 @@ RUN /usr/bin/pip3 install nglview==2.7.7
 RUN /usr/local/bin/jupyter nbextension enable nglview --py --sys-prefix
 
 # Install and enable appmode.
-WORKDIR /opt/
```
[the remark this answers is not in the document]
yakutovicha: Yes, I want to put the `requirements.txt` file to the `/opt` folder. This is why I moved the change of the working directory before installing all the requirements.

README.md
```diff
@@ -28,6 +28,23 @@ Where `PORT` is any free port on your machine (typically it is 8888) and `PATH_T
...
+
+First make sure you have python 3.7 available in your system.
+If that is the case, then adjust the [`Pipfile`](Pipfile) according to the latest releases.
+Then do:
+```
+pip install pipenv # If it is already installed, make sure it is the latest version.
+pipenv lock --python 3.7 # To generate Pipfile.lock (takes 1-2 minutes).
+pipenv lock --requirements > requirements.txt_stable # To extract the dependencies of the stable packages.
+pipenv lock --dev --requirements > requirements.txt_dev # To extract the dependencies of the development packages.
+cat requirements.txt_stable requirements.txt_dev | sort | uniq > requirements.txt
```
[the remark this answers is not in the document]
yakutovicha: I thought about this, and I actually prefer to keep it. I can modify the text saying something like "if you want to include the `[dev]` requirements, follow this path, otherwise just do `pipenv lock --requirements > requirements.txt`". 

The point is: if next time I will have a package with weird dependencies (say, aiida is set to be lower than 1.5.0, even though it is actually working) I don't want to search again for the information on how to handle it.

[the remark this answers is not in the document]
csadorf: Yes, would much prefer to keep those as "additional instructions", but not use or advocate this approach as the "standard way".

Dockerfile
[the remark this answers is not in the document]
csadorf: Ah, missed line 59, that's fine then.

.gitignore
csadorf: I'd prefer to remove those "extra" requirements from the .gitignore file.

yakutovicha: It is ok. However, I would still prefer to keep the lock file.

csadorf: Yes, that's fine.

README.md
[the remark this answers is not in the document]
yakutovicha: I changed my mind. Soon pip won't allow us to create such inconsistent environments, therefore it makes total sense to be more strict here.

If a package wouldn't fit the environment, we simply won't include it.

README.md
```diff
@@ -28,6 +28,21 @@ Where `PORT` is any free port on your machine (typically it is 8888) and `PATH_T
...
 
+# Update requirements.txt
+
+First make sure you have python 3.7 available in your system.
+If that is the case, then adjust the [`Pipfile`](Pipfile) according to the latest releases.
+Then do:
+```
+pip install pipenv # If it is already installed, make sure it is the latest version.
+pipenv lock --python 3.7 # To generate Pipfile.lock (takes 1-2 minutes).
+pipenv lock --requirements > requirements.txt
```
csadorf, line 39: Could this be combined to:
```suggestion
pipenv lock --python 3.7 --requirements > requirements.txt
```
?

yakutovicha: let me try.

yakutovicha: yes, works.

csadorf (approved):

PR #138, opened 2020-11-06
Release 20.11.0

yakutovicha:


csadorf:
@yakutovicha I think we should consider disabling the Travis CI.

yakutovicha:
> @yakutovicha I think we should consider disabling the Travis CI.

I agree - it is getting really annoying. Would you like to make a PR for that?

csadorf:
> > @yakutovicha I think we should consider disabling the Travis CI.
> 
> I agree - it is getting really annoying. Would you like to make a PR for that?

Sure.

PR #139, opened 2020-11-06
Install openbabel from conda-forge

### LINKED ISSUE #129: Update rdkit and openbabel verisons
To have a more robust implementation of the SMILES converter
conversion should be first tried via rdkit via a try---except
and then via openbabel.

For some strange reason the present versions of openbabel and rdkit interfer one each other
it should be better to install:

rdkit=2020.03.6
openbabel=3.1.1

https://github.com/aiidalab/aiidalab-widgets-base/blob/cce7876ab87b3b5876683c4a71c62dfc3117fbd9/aiidalab_widgets_base/structures.py#L517

a correct installation should allow to process the following sequence:
`from openbabel import pybel as pb
from openbabel import openbabel as ob
from rdkit.Chem import MolFromSmiles,MolToMolFile
from rdkit import Chem
from rdkit.Chem import AllChem
#some pybel code
#end some pybel code

m = Chem.MolFromSmiles("C")
m = Chem.AddHs(m)
    
AllChem.EmbedMolecule(m, maxAttempts=20, randomSeed=42)
AllChem.UFFOptimizeMolecule(m,maxIters=100)
m.GetConformer().GetPositions()`

the following sequence instead (with present way to upload pubel) crashes:

`import pybel as pb
import openbabel as ob
from rdkit.Chem import MolFromSmiles,MolToMolFile
from rdkit import Chem
from rdkit.Chem import AllChem
#some pybel code
#end some pybel code

m = Chem.MolFromSmiles("C")
m = Chem.AddHs(m)
    
AllChem.EmbedMolecule(m, maxAttempts=20, randomSeed=42)
AllChem.UFFOptimizeMolecule(m,maxIters=100)
m.GetConformer().GetPositions()`

yakutovicha:
fixes #129

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -102,8 +102,7 @@ RUN git clone https://github.com/aiidalab/aiidalab-optimade.git && cd aiidalab-o
 RUN pip install -e ./aiidalab-optimade
 
 # Install some useful packages that are not available on PyPi
-RUN conda install --yes -c conda-forge rdkit
-RUN conda install --yes -c openbabel openbabel
+RUN conda install --yes -c conda-forge rdkit openbabel
 
 # Prepare user's folders for AiiDAlab launch.
 COPY opt/aiidalab-singleuser /opt/
--- a/Dockerfile
+++ b/Dockerfile
@@ -102,7 +102,10 @@ RUN git clone https://github.com/aiidalab/aiidalab-optimade.git && cd aiidalab-o
 RUN pip install -e ./aiidalab-optimade
 
 # Install some useful packages that are not available on PyPi
-RUN conda install --yes -c conda-forge rdkit openbabel
+RUN conda install --yes -c conda-forge \
+  openbabel \
+  rdkit \
+  && conda clean --all
 
 # Prepare user's folders for AiiDAlab launch.
 COPY opt/aiidalab-singleuser /opt/


Dockerfile
```diff
@@ -102,8 +102,7 @@ RUN git clone https://github.com/aiidalab/aiidalab-optimade.git && cd aiidalab-o
 RUN pip install -e ./aiidalab-optimade
 
 # Install some useful packages that are not available on PyPi
-RUN conda install --yes -c conda-forge rdkit
-RUN conda install --yes -c openbabel openbabel
+RUN conda install --yes -c conda-forge rdkit openbabel
```
csadorf, line 105: I think we should structure the `conda install` commands like [`apt install` commands](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#sort-multi-line-arguments):
```suggestion
RUN conda install --yes -c conda-forge \
  openbabel \
  rdkit \
  && conda clean --all
```

csadorf (changes requested):

Dockerfile
csadorf: Additionally we might want to add comments as to why we are installing some of these packages.

yakutovicha: ok, sure

yakutovicha: Is it possible to add a comment after the `\`  symbol? It didn't work for me.

csadorf: Ah, didn't realize that. Then just in the commit message, because it will be visible via `git blame`.

csadorf (approved):
Please add the reasons for adding those packages in the commit message, thx.

Dockerfile
yakutovicha: ok

PR #141, opened 2020-11-06
Move remaining dependencies from aiidalab to Pipfile

yakutovicha:


csadorf (approved):
Looks fine, although I'd prefer if we could add the reason for adding a specific dependency with an inline comment.

PR #143, opened 2020-11-25
Pin the aiidalab-widgets base version.

csadorf:
Otherwise a push to the aiidalab-widgets-base repository has potential
to break new aiidalab user instances.


yakutovicha (commented):
I am not entirely sure we really need that. We merge things to the master branch only when making new releases. If we introduce this change, any update will require to "manually" change the version. Even if we forget to do that, the users will be notified by AiiDAlab that update is available and users will do it by themselves - so the end result is the same.

csadorf:
> I am not entirely sure we really need that. We merge things to the master branch only when making new releases. If we introduce this change, any update will require to "manually" change the version. Even if we forget to do that, the users will be notified by AiiDAlab that update is available and users will do it by themselves - so the end result is the same as the status quo.

Ok, but the latest update for example broke things. So we need to be more careful about merging into master then.

csadorf:
I am proposing this change once more, because issues like #147 clearly demonstrate that having ill-defined dependencies in our environment build and setup flow poses a huge risk that the environment creation is not reproducible and thus vulnerable to unexpectedly break at undetermined (and highly inopportune) times. **All** updates to the environment should be made in a controlled manner.

yakutovicha (approved):
agreed!

PR #154, opened 2020-12-14
Install and enable ipywidgets-extended.

csadorf:
Resolves https://github.com/CasperWA/voila-optimade-client/issues/235 .


CasperWA (approved):
_This_ is what I needed :)

The long-term solution here, would be for me to contribute this widget back to `ipywidgets` and then this can be removed again?

csadorf:
> _This_ is what I needed :)
> 
> The long-term solution here, would be for me to contribute this widget back to `ipywidgets` and then this can be removed again?

Yes, and we should also think about implementing a more systematic way to install and enable required Jupyter extensions.

CasperWA:
@yakutovicha I think we're waiting on your approval and merging here :)

yakutovicha (approved):
approved :)

PR #174, opened 2021-04-12
Update dependencies

yakutovicha:


Pipfile
```diff
@@ -6,7 +6,7 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-aiidalab = "~=20.11.0b0"
+aiidalab = "~=21.04"
```
csadorf, line 9:
```suggestion
aiidalab = "~=21.04.0"
```
would probably be better

csadorf (changes requested):
Just for the record, we'll have to fix the incompatibilities in the aiidalab package so that the home app tests pass before we can merge this.

yakutovicha:
> Just for the record, we'll have to fix the incompatibilities in the aiidalab package so that the home app tests pass before we can merge this.

definitely, I am now going to proceed with a new release of aiidalab.

Pipfile
yakutovicha: This makes me think that we should indeed switch to the semantic versioning, as the `aiidalab` package is not a meta-package anymore.

csadorf: I don't think we need semantic versioning, but we could use a mixed approach where at least the point-releases should be compatible. But in principle we should not use the comaptible-release operator here, since there is no assumption about compatibility between releases. So this would actually be more appropriate:
```suggestion
aiidalab = "==21.04.01"
```

yakutovicha: it is `21.04.1`, but I agree with your comment.

yakutovicha:
also tested on a locally build image - everything seems to be ok

csadorf (approved):

PR #177, opened 2021-05-05
Install Quantum ESPRESSO (QE) via conda (conda-forge).

csadorf:
QE is now available on conda-forge. This changeset takes advantage of that and install QE directly via conda instead of apt.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,6 @@ RUN apt-get update && apt-get install -y  \
     python3-pip           \
     python3-setuptools    \
     python3-wheel         \
-    quantum-espresso      \
   && rm -rf /var/lib/apt/lists/*
 
 # Install what is needed for Jupyter Lab.
@@ -95,6 +94,7 @@ RUN /usr/bin/pip3 install \
 # The 2020.09.2 version of rdkit introduced an implicit dependency on tornado>=6.
 RUN conda install --yes -c conda-forge \
   openbabel==3.1.1 \
+  qe==6.7.0 \
   rdkit==2020.09.1 \
   && conda clean --all
 


csadorf:
Please note that the QE app tests pass, however that is because they do not go through the setup-code and run calculation path. In fact the setup-code path should fail, because the default `pw.x` path is hard-coded to be `/usr/bin/pw.x` within the QE app. This does not mean that the QE app is directly broken, but when a user clicks on "setup code", the pre-filled values will be wrong.

ltalirz:
I think it's healthy to keep things simple if possible (i.e. to stick with one package index for the moment).

I guess the installation of QE via apt was not explicitly recorded before (?), so you're at least not making things worse by switching to conda here.
If you start running into issues at some point (say, if you need to install multiple conda-packages with conflicting dependencies) you can revisit it.

In any case, please take this just as a gut feeling since I'm not too familiar with the setup you've chosen.

csadorf:
> I guess the installation of QE via apt was not explicitly recorded before (?), so you're at least not making things worse by switching to conda here.

Yes, however your answer made me realize that I forgot that QE is not technically a "real dependency" of the app. Since running the workchain on the localhost is only one option to run the app. We could consider it an extra if you will.

yakutovicha:
btw, cp2k should probably also be installed through conda. It is also available there.

yakutovicha:
As we agreed to keep in this package only the core AiiDAlab functionality - the installation of QE/CP2K has been moved to [aiidalab-docker-stack-demo](https://github.com/aiidalab/aiidalab-docker-stack-demo/blob/develop/Dockerfile#L7-L10).

PR #181, opened 2021-06-04
Update aiidalab, aiidalab-qe, temporarily remove aiida-siesta plugin.

yakutovicha:


requirements.txt
```diff
@@ -15,7 +15,7 @@ aiida-optimize~=0.4
 aiida-qeq~=1.0.0a1
 aiida-quantumespresso~=3.4
 aiida-raspa~=1.2
-aiida-siesta~=1.1
+#aiida-siesta~=1.1
```
csadorf, line 18: If we comment it out, then we should also provide a reason, otherwise just remove it.

yakutovicha: removed.

csadorf (approved):
Looks fine to me, however I did not test it.

yakutovicha:
> Looks fine to me, however I did not test it.

ok, thanks. The image will remain in the dev instance for testing for one week.

PR #184, opened 2021-06-08
Deploy clean AiiDAlab container.

### LINKED ISSUE #175: Possibility to deploy a clean AiiDAlab
Right now, e.g. QE is installed by default and that seems rather hard coded. We should open up for an easy way to avoid this and just deploy the basics of AiiDAlab.

In my view this would also not include material science related widgets, only the computer, code etc.

yakutovicha:
fixes #175


csadorf:
@yakutovicha Is this ready for review?

Dockerfile
```diff
@@ -1,26 +1,27 @@
...
 
-LABEL maintainer="Materials Cloud Team <aiidalab@materialscloud.org>"
+LABEL maintainer="AiiDAlab Team <aiidalab@materialscloud.org>"
 
 # Configure environment.
 ENV AIIDALAB_HOME /home/${SYSTEM_USER}
 ENV AIIDALAB_APPS ${AIIDALAB_HOME}/apps
 ENV AIIDALAB_DEFAULT_GIT_BRANCH master
 
+ENV QE_APP_VERSION develop
```
csadorf, line 10: This should be empty, so that the base image does not install the app:
```suggestion
ENV QE_APP_VERSION
```

Dockerfile
```diff
@@ -1,26 +1,27 @@
...
+ENV AWB_APP_VERSION v1.0.0b18
```
csadorf, line 11: This should be empty so that the base image does not install the app:
```suggestion
ENV AWB_APP_VERSION
```

opt/prepare-aiidalab.sh
```diff
@@ -64,15 +64,16 @@ if [[ ${INITIAL_SETUP} == true ||  "${AIIDALAB_SETUP}" == "true" ]]; then
   if [ ! -e /home/${SYSTEM_USER}/apps/aiidalab-widgets-base ]; then
```
csadorf, line 67:
```suggestion
if [ ! -z ${AWB_APP_VERSION} ] && [ ! -e /home/${SYSTEM_USER}/apps/aiidalab-widgets-base ]; then
```

opt/prepare-aiidalab.sh
```diff
@@ -64,15 +64,16 @@ if [[ ${INITIAL_SETUP} == true ||  "${AIIDALAB_SETUP}" == "true" ]]; then
...
-    git checkout ${AIIDALAB_DEFAULT_GIT_BRANCH}
-    git reset --hard v1.0.0b18
+    git checkout ${AWB_APP_VERSION}
     cd -
-  fi 
-  # Quantum Espresso app.
+  fi
+  
+ # Quantum Espresso app.
   if [ ! -e /home/${SYSTEM_USER}/apps/quantum-espresso ]; then
```
csadorf, line 75:
```suggestion
  if [ ! -z ${QE_APP_VERSION} ] && [ ! -e /home/${SYSTEM_USER}/apps/quantum-espresso ]; then
```

opt/prepare-aiidalab.sh
```diff
@@ -64,15 +64,16 @@ if [[ ${INITIAL_SETUP} == true ||  "${AIIDALAB_SETUP}" == "true" ]]; then
...
     git clone https://github.com/aiidalab/aiidalab-qe.git /home/${SYSTEM_USER}/apps/quantum-espresso
     cd /home/${SYSTEM_USER}/apps/quantum-espresso
-    git reset --hard v20.12.0
+    git checkout ${QE_APP_VERSION}
```
csadorf, line 78:
```suggestion
    git checkout "${QE_APP_VERSION}"
```

Just in case it every contains spaces...

opt/prepare-aiidalab.sh
```diff
@@ -64,15 +64,16 @@ if [[ ${INITIAL_SETUP} == true ||  "${AIIDALAB_SETUP}" == "true" ]]; then
   if [ ! -e /home/${SYSTEM_USER}/apps/aiidalab-widgets-base ]; then
     git clone https://github.com/aiidalab/aiidalab-widgets-base /home/${SYSTEM_USER}/apps/aiidalab-widgets-base
     cd /home/${SYSTEM_USER}/apps/aiidalab-widgets-base
-    git checkout ${AIIDALAB_DEFAULT_GIT_BRANCH}
-    git reset --hard v1.0.0b18
+    git checkout ${AWB_APP_VERSION}
```
csadorf, line 70:
```suggestion
    git checkout "${AWB_APP_VERSION}"
```

csadorf (changes requested):

opt/prepare-aiidalab.sh
csadorf: Just realized, this missing the dependency installation, is it not?

yakutovicha: > Just realized, this is missing the dependency installation, is it not?

That depends on whether we want to keep the `aiidalab-widgets-base` **package** linked to the image version.

If we link it - then it is listed as the dependency in the `requirements.txt` file. If we don't - then we install the dependencies here.

my_init.d/factory_reset.sh
```diff
@@ -0,0 +1,27 @@
...
+fi
+
+if [[ "${AIIDALAB_FACTORY_RESET}" == 1 ]]; then
+  rm -r /home/${SYSTEM_USER}/apps
+  rm -r /home/${SYSTEM_USER}/.local
+
+elif [[ "${AIIDALAB_FACTORY_RESET}" == 2 ]]; then
+  rm -r /home/${SYSTEM_USER}/*
+
+fi
```
csadorf, line 25: This would be a prime use case for [case statements](https://linuxize.com/post/bash-case-statement/).

But this is only relevant if we don't adopt the bit mask approach I recommended in an earlier comment.

csadorf (changes requested):

opt/prepare-aiidalab.sh
csadorf: I'm not sure what you mean by this. The AWB version is now "linked" through the AWB_APP_VERSION env variable, so it should be installed here.

csadorf: Ok, I think I understand better now what you mean, but that would be very non-intuitive if we install a different package version somewhere else, don't you think?

my_init.d/factory_reset.sh
yakutovicha: We discussed elsewhere and agreed to take the `case statements` approach for the moment. In case of a need, we can switch to the mask approach later.

opt/prepare-aiidalab.sh
yakutovicha: Agreed: I added the pip install.

my_init.d/factory_reset.sh
```diff
@@ -0,0 +1,34 @@
...
+    exit 0
+    ;;
+
+  1)
+    rm -rf "/home/${SYSTEM_USER}/apps"
+    rm -rf "/home/${SYSTEM_USER}/.local"
+    ;;
+
+  2)
+    rm -rf /home/${SYSTEM_USER}/*
```
csadorf, line 27: The current implementation will not delete hidden directories.

This should work (from [here](https://unix.stackexchange.com/a/77128)):
```suggestion
    find home/${SYSTEM_USER}/ -mindepth 1 -delete
```

requirements.txt
```diff
@@ -1,25 +1,4 @@
...
-aiida-nanotech-empa~=0.2
-aiida-optimize~=0.4
-aiida-qeq~=1.0.0a1
-aiida-quantumespresso~=3.4
-aiida-raspa~=1.2
-aiida-tbmodels~=0.3
-aiida-vasp~=2.1
-aiida-wannier90~=2.0
-aiida-yambo~=1.3
-aiida-zeopp~=1.1
```
csadorf, line 22: I don't think we should install any of the dependencies left after this line. They should be dependencies of AWB and the QE app, no?

csadorf (changes requested):

requirements.txt
yakutovicha: Well, yes, let's remove it and test the apps. In case things are missing - we will put them as the dependencies of the apps.

yakutovicha: I changed my mind. If we do that right now - we end up having a non-functioning image. I propose that we distribute the remaining dependencies in a separate PR.

csadorf: Why would it be non-functioning?

csadorf (approved):
Overall I think this should be fine now. However, I realized that for the factory reset we should probably use a file `~/AIIDALAB_FACTORY_RESET`, because otherwise it would be very difficult to trigger the reset from within an AiiDAlab instance. I am ok with adding that later though.

requirements.txt
yakutovicha: because dependencies like markdown are not set anywhere.

yakutovicha: we just need to distribute the remaining dependencies to the appropriate packages.

PR #186, opened 2021-06-10
Feature/factory reset

csadorf:


my_init.d/factory_reset.sh
```diff
@@ -1,34 +1,50 @@
-#!/bin/bash -e
-
-# Debugging.
-set -x
```
yakutovicha, line 4: Are you sure you want to delete that? I think, if something goes wrong - the image should not start.

my_init.d/factory_reset.sh
```diff
@@ -1,34 +1,50 @@
...
+if (( (${RESET_MODE} & 0x2) == 0x2)); then
+  echo "Factory reset: Remove user home directory contents."
+  find /home/${SYSTEM_USER}/ -mindepth 1 -delete
+fi
+
+# mode & 100 -> delete the factory reset file (if it exists)
+if (( (${RESET_MODE} & 0x4) == 0x4)); then
+  echo "Factory reset: Removing factory reset file (~/AIIDALAB_FACTORY_RESET) (if it exists)."
+  rm -f "${HOME}/AIIDALAB_FACTORY_RESET"
+fi
```
yakutovicha, line 50: If there reset mode is unknown, shouldn't the image fail to start?

my_init.d/factory_reset.sh
```diff
@@ -1,34 +1,50 @@
...
+if (( (${RESET_MODE} & 0x1) == 0x1)); then
+  echo "Factory reset: Remove apps (~/apps/) and local software installation (~/.local/)."
+  rm -rf "/home/${SYSTEM_USER}/apps"
+  rm -rf "/home/${SYSTEM_USER}/.local"
+fi
+
+# mode & 010 -> delete all home directory contents
+if (( (${RESET_MODE} & 0x2) == 0x2)); then
+  echo "Factory reset: Remove user home directory contents."
+  find /home/${SYSTEM_USER}/ -mindepth 1 -delete
```
yakutovicha, line 43:
```suggestion
  find "/home/${SYSTEM_USER}/" -mindepth 1 -delete
```
I believe this should work, but please test.

yakutovicha (commented):
thanks, @csadorf: a few questions/suggestions.

my_init.d/factory_reset.sh
csadorf: The [`-x` flag](https://stackoverflow.com/a/36273740) only instructs bash to echo all commands, it does not influence whether the script fails on error. The `-e` flag is still set (see line 1).

my_init.d/factory_reset.sh
csadorf: No, I think it would be more robust to print a warning if anything.

yakutovicha: Fine with me. Please add a warning then.

my_init.d/factory_reset.sh
csadorf: Done.

my_init.d/factory_reset.sh
```diff
@@ -1,34 +1,43 @@
...
+# Bit mask for performing factory reset of the AiiDAlab environment:
+# 0b000 0 - No reset (noop).
+# 0b001 1 - Remove locally installed software and apps (removes ~/apps/ and ~/.local/).
+# 0b010 2 - Remove all files and directories within the users home directory.
+
+# If the ~/AIIDALAB_FACTORY_RESET file exists, parse it (and remove it).
+if [ -e "${HOME}/AIIDALAB_FACTORY_RESET" ]; then
+  RESET_MODE="`cat ${HOME}/AIIDALAB_FACTORY_RESET`"
+  rm -f "${HOME}/AIIDALAB_FACTORY_RESET"  # Remove file after parsing.
+fi
```
yakutovicha, line 15:
```suggestion
# If the ~/AIIDALAB_FACTORY_RESET file exists, parse it (and remove it).
if [ -e "/home/${SYSTEM_USER}/AIIDALAB_FACTORY_RESET" ]; then
  RESET_MODE="`cat /home/${SYSTEM_USER}/AIIDALAB_FACTORY_RESET`"
  rm -f "/home/${SYSTEM_USER}/AIIDALAB_FACTORY_RESET"  # Remove file after parsing.
fi
```
The scripts from `my_init.d` folder are run by root, so using `${HOME}` doesn't make sense.

yakutovicha (changes requested):
The container didn't actually start for me, see the suggested change.

csadorf:
@yakutovicha I tested this locally and it appears to work as expected.

yakutovicha (approved):
Thanks, @csadorf! Please go ahead with merging.

PR #189, opened 2021-06-30
The AIIDALAB_DEFAULT_APPS variable is whitespace delimited.

csadorf:
Specifying newline-delimited variables is typically not well supported
in various contexts and iterating over a whitespace delimited variable
is also easier.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,15 +11,19 @@ ENV AIIDALAB_APPS ${AIIDALAB_HOME}/apps
 ENV AIIDALAB_DEFAULT_GIT_BRANCH master
 
 # Specify which apps to install in addition to the home app. The
-# AIIDALAB_DEFAULT_APPS variable should be a multiline variable where each line
-# has the following format:
-# <app name>@<git url>@<version>
+# AIIDALAB_DEFAULT_APPS variable should be a whitespace-delimited variable
+# where each entry has the following format: <app name>@<git url>@<version>
 # The version should be a reference that can be checked out via
 # 'git checkout <version>'.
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -11,14 +11,17 @@ ENV AIIDALAB_APPS ${AIIDALAB_HOME}/apps
...
-# Example for a properly formatted AIIDALAB_DEFAULT_APPS variable:
-#   aiidalab-widgets-base@https://github.com/aiidalab/aiidalab-widgets-base@v1.0
-#   quantum-espresso@https://github.com/aiidalab/aiidalab-qe@v1.1
+#
+# Example for setting the AIIDALAB_DEFAULT_APPS variable:
+#
+#   AIIDALAB_DEFAULT_APPS+="aiidalab-widgets-base@https://github.com/aiidalab/aiidalab-widgets-base@v1.0 "
+#   AIIDALAB_DEFAULT_APPS+="quantum-espresso@https://github.com/aiidalab/aiidalab-qe@v1.1 "
+#
+# Please not the space at the end to ensure that entries are whitespace delimited.
```
yakutovicha, line 24:
```suggestion
# Please do not put space at the end to ensure that entries are whitespace delimited.
```
unless I got it wrong.

yakutovicha (commented):
Thanks, @csadorf, only a small comment.

Dockerfile
csadorf: It should be:
```suggestion
# Please note the space at the end to ensure that entries are whitespace delimited.
```
However, I found placing the whitespace at the beginning is actually even better. I'll push the change.

yakutovicha (approved):
Looks good! Thanks, @csadorf!

PR #197, opened 2021-09-13
bump aiida-core baseimage

ltalirz:
* bump aiida-core baseimage
* adapt to ubuntu 20.04 environment
* move out all python dependencies for the jupyter notebook server environment into a separate file
* add workflow to run on every commit
* name docker container in `run.sh` script

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,6 +39,9 @@ RUN apt-get update && apt-get install -y  \
     libssl-dev            \
     libffi-dev            \
     povray                \
+    python3-pip           \
+    python3-setuptools    \
+    python3-wheel         \
   && rm -rf /var/lib/apt/lists/*
 
 # Dependencies needed for Jupyter Lab.
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,8 +40,6 @@ RUN apt-get update && apt-get install -y  \
     libffi-dev            \
     povray                \
     python3-pip           \
-    python3-setuptools    \
-    python3-wheel         \
   && rm -rf /var/lib/apt/lists/*
 
 # Dependencies needed for Jupyter Lab.
@@ -57,11 +55,14 @@ RUN wget --quiet -P /tmp/ \
   && mv ./ngrok /usr/local/bin/ \
   && rm -f /tmp/ngrok-2.3.35-linux-amd64.zip
 
+# Get recent version of pip (needed for `pip cache` command).
+# New pip executable is installed into /usr/local/bin
+RUN /usr/bin/pip install --upgrade pip
+
 # Jupyter dependencies installed into system python environment.
-RUN /usr/bin/python3 -m pip install -U pip
 COPY requirements-server.txt .
-RUN /usr/bin/pip3 install -r /opt/requirements-server.txt \ 
-    && /usr/bin/pip3 cache purge
+RUN /usr/local/bin/pip install -r /opt/requirements-server.txt \ 
+    && /usr/local/bin/pip cache purge
 
 # Enable server extensions
 RUN /usr/local/bin/jupyter serverextension enable --py --sys-prefix nbserverproxy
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -27,41 +27,53 @@ ENV AIIDALAB_DEFAULT_GIT_BRANCH master
 ENV AIIDALAB_DEFAULT_APPS ""
 
 USER root
+WORKDIR /opt/
 
 # Install OS dependencies.
+# libssl-dev needed for ?
```
ltalirz, line 33: do we know?

Dockerfile
```diff
@@ -27,41 +27,53 @@ ENV AIIDALAB_DEFAULT_GIT_BRANCH master
...
-RUN /usr/bin/python3 -m pip install -U pip
-
-# Install Jupyter-related things in the root environment.
-RUN /usr/bin/pip3 install          \
-    'jupyterhub==1.4.0'            \
-    'jupyterlab==3.0.14'            \
-    'notebook==6.3.0'
-
-# Install ngrok to be able to proxy AiiDA RESTful API server.
+# Install ngrok to be able to proxy AiiDA RESTful API server. ### is this used in the home app?
```
ltalirz, line 51: is this currently needed by the "base"  aiidalab?

requirements-server.txt
```diff
@@ -0,0 +1,15 @@
+# Dependencies for notebook server and interaction with JupyterHub
+jupyterhub==1.4.0
+jupyterlab==3.0.14
+notebook==6.3.0
+# Dependencies for jupyter widgets
```
ltalirz, line 5: do we need all of these?

requirements-server.txt
```diff
@@ -0,0 +1,15 @@
...
+bqplot==0.12.25
+ipytree==0.1.8
+ipywidgets-extended==1.0.5
+nglview==2.7.7
+widget-periodictable==2.1.5
+# nbserverproxy. ### are we using this?
+nbserverproxy==0.8.8
+# Install voila package and AiiDAlab voila template. ### another candidate for removal? is this documented?
```
ltalirz, line 13: what about voila - is this being used?

requirements-server.txt
```diff
@@ -0,0 +1,15 @@
...
+jupyterhub==1.4.0
+jupyterlab==3.0.14
+notebook==6.3.0
+# Dependencies for jupyter widgets
+bqplot==0.12.25
+ipytree==0.1.8
+ipywidgets-extended==1.0.5
+nglview==2.7.7
+widget-periodictable==2.1.5
+# nbserverproxy. ### are we using this?
```
ltalirz, line 11: nbserverproxy required?

requirements-server.txt
yakutovicha: I think we haven't made a final decision on whether to use it or not. Most probably, it will be used in Jupyter Lab. So for the moment, I would leave it as is.

requirements-server.txt
yakutovicha: Unfortunately yes. They are here just for activation purposes.

requirements-server.txt
yakutovicha: Probably not, I don't remember what was the reason for it to be added. Additionally, `nbserverproxy` is an old name. Now the package is called [`jupyter-server-proxy`](https://github.com/jupyterhub/jupyter-server-proxy).

Dockerfile
yakutovicha: @csadorf, can you please reply to this?

Dockerfile
yakutovicha: Not really, those were added in the initial commit: https://github.com/aiidalab/aiidalab-docker-stack/commit/470ba8f681e79cf6da94b26b3291ef34b19d6e25

requirements-server.txt
yakutovicha: I think the intention was to give access to the AiiDA REST API: cd2e1390edf62730f3b5717a9e83121710c50b0a

Dockerfile
ltalirz: ok, I'll add a comment to that effect but will leave them in for the moment in order not to lead to any unforeseen issues

Dockerfile
csadorf: Ngrok was added for in the context of an AiiDA tutorial to make it significantly easier to reverse-proxy the AiiDA web server. It is not used by the home app. I would recommend to keep it.

ltalirz: Alright!

ltalirz:
@yakutovicha This was my last question; from my side the PR is good to go. Please proceed with the review

yakutovicha (approved):
LGTM, thanks @ltalirz!

Issue #183, opened 2021-06-05
List of hindrances when starting to develop


mbercx:
When starting to develop after running:

```
./run.sh 8888 ~/aiidalab
```

I run into a list of hindrances before I can really get to work. I'll keep track of them here. Not sure if these can be "fixed", but the steps to resolve them an get to work should be documented. Or a different pathway to starting to develop should be explained or easy to find.

- [x] Although the `apps` are installed in the `~/aiidalab` directory and hence easily accessible, you still have to change the ownership to your user before you can start adapting them or using e.g. git commands:
   ```
  (qeapp) mbercx@theospc46:~/aiidalab$ ls -l
  total 4
  drwxrwxr-x 5 administrator administrator 4096 Jun  5 09:49 apps
  ```
  Once I run:
  ```
  sudo chown -R mbercx:administrator apps/
  ```
  I'm in business! Maybe we can change the ownership in the setup?
- [x] For future releases that ship with the `aiidalab-qe-workchain`: immediately install this package in editable mode. This is really just for yours truly, but would be nice if I could skip the step to have to log in as root and do this. 😁 
- [x] Although upon initial deployment, the link+token is outputted so I can get to work right away, when I restart my computer for example I need to give the token again when I log in. If I hadn't kept this somewhere, I wouldn't really know how to proceed, except for restart the deployment.
- [x] The docker container now has a default generated name like `strange_dewdney` or `heuristic_wiles`. Is there a reason not to specify this name?

PS: I thought there were more when I first started opening this issue, but seems I just messed up a lot when I last set up this dev environment. 😅  Will still add anything I can think of to make the dev process smoother here.

csadorf:
>  Although the apps are installed in the ~/aiidalab directory and hence easily accessible, you still have to change the ownership to your user before you can start adapting them or using e.g. git commands: 

The `~/aiidalab` directory (or whatever path you choose) is a bind mount and docker changes the ownership. It might be possible to prevent that, but I would need to look into how.

> For future releases that ship with the aiidalab-qe-workchain: immediately install this package in editable mode. This is really just for yours truly, but would be nice if I could skip the step to have to log in as root and do this.

I don't think there is a way around this, since for users of the app, the work chain must be installed independent of the app. Otherwise any update or re-installation of the app would break the environment.

ltalirz:
docker container name fixed in 3977b7632775358a7239e146608921e1ce41d863 (name is `aiidalab`)

[...]

PR #203, opened 2021-09-15
Remove `CONTARINER_NAME` variable from run.sh

yakutovicha:
The containers must always have different name, otherwise they
wouldn't start.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -153,13 +153,12 @@ if [ "${_arg_build}" == "on" ]; then
 fi
 
 echo "Launching the container..."
-CONTAINER_NAME=aiidalab
-CONTAINER_ID=`docker run -d -p "${_arg_port}:8888" -e JUPYTER_TOKEN="${_arg_token}" -v "${_arg_home_dir}:/home/aiida" --name ${CONTAINER_NAME} "${_arg_image}"`
+CONTAINER_ID=`docker run -d -p "${_arg_port}:8888" -e JUPYTER_TOKEN="${_arg_token}" -v "${_arg_home_dir}:/home/aiida" "${_arg_image}"`
 
-echo "Waiting for container '${CONTAINER_NAME}' to start..."
+echo "Waiting for container to start..."
 docker exec --tty ${CONTAINER_ID} wait-for-services
 
-echo "Container '${CONTAINER_NAME}' started successfully."
+echo "Container started successfully."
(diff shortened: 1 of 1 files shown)


ltalirz:
This was done in response to a request by @mbercx , see https://github.com/aiidalab/aiidalab-docker-stack/issues/183

In his use case, I guess only one container is ever needed at a given time.

Is the `./run.sh` also used by the jupyterhub installations?

Perhaps we could add a `--name` argument to the runscript that names the container only if specified.

csadorf (changes requested):
The use case of starting multiple containers appears less safe and likely to me than explicitly setting a different name when desired. What happens when both containers mount the same home directory?

I would instead suggest to make the name an optional parameters which a user can set to a different value when desired, e.g., when they want to start more than one container.

yakutovicha:
> The use case of starting multiple containers appears less safe and likely to me than explicitly setting a different name when desired

I think I didn't explain the issue properly. The thing is: even if the container with "aiidalab" name is stopped, starting another one with the same name wouldn't be possible.

ltalirz:
> I think I didn't explain the issue properly. The thing is: even if the container with "aiidalab" name is stopped, starting another one with the same name wouldn't be possible.

You just explain to people that they remove the container with `docker rm -f aiidalab` , see https://github.com/aiidalab/aiidalab/pull/214

csadorf:
As I had pointed out before, I would suggest that we adopt the aiidalab-docker-stack-demo model with a docker-compose.yml and control script. It's more easily configurable and avoids a lot of these issues or at least makes them easier to deal with.

ltalirz:
I've had a look at the repo docker-stack-demo and the docker-compose approach looks good to me.

Ideally, we could maintain that here in the docker-stack repo; switching to the docker-stack-demo should then just be changing one variable.

yakutovicha:
Thanks to both. I agree that we should probably adopt the docker-compose approach here. I also agree that when running multiple containers at the same time user should be requested to provide a new name explicitly.

The only thing I would like to comment on is this statement:

> You should only stop the container if you plan restarting it.

I sometimes stop the containers when the machine is overloaded or I just switch off my computer. The only thing I would like to avoid is a message like "Sorry, I cannot start the 'aiidalab' container" because it already exists.

yakutovicha:
superseded by #217

PR #211, opened 2021-09-30
Uninstall aiidalab package from user site if present.

csadorf:
To prevent that it interferes with the system package.

Related to https://github.com/aiidalab/aiidalab/issues/220 .

### DIFF
--- a/opt/prepare-aiidalab.sh
+++ b/opt/prepare-aiidalab.sh
@@ -96,3 +96,11 @@ find -L /home/${SYSTEM_USER} -maxdepth 3 -name apps_meta.sqlite -delete
 
 # Remove old temporary notebook files.
 find -L /home/${SYSTEM_USER}/apps -maxdepth 2 -type f -name .*.ipynb -delete
+
+# Uninstall aiidalab from user packages (if present).
+# Would otherwise interfere with the system package.
+USER_AIIDALAB_PACKAGE="$(/opt/conda/bin/python -c 'import site; print(site.USER_SITE)')/aiidalab"
+if [ -e ${USER_AIIDALAB_PACKAGE} ]; then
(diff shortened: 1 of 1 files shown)


yakutovicha:
@csadorf, can you rebase the branch on top of develop, please? Currently, it requires aiida-core:1.6.5

csadorf:
> @csadorf, can you rebase the branch on top of develop, please? Currently, it requires aiida-core:1.6.5

Sorry about that, done.

yakutovicha (commented):
Ok, the proposed change worked for me.

This means if a package was installed in the user's environment - everything worked.

At the same time, if the use of the `aiidalab` package was configured by the `aiidalab-dev` tool - it won't be overritten. Is it what we want? Or do we maybe always want to update the link to the home package at the container startup?

csadorf:
> At the same time, if the use of the `aiidalab` package was configured by the `aiidalab-dev` tool - it won't be overritten. Is it what we want? Or do we maybe always want to update the link to the home package at the container startup?

I would think that this is a very specific case in which one would hope that the user (developer) knows what they are doing. Here I am merely trying to prevent the current issue, that the aiidalab package was inadvertently installed into the user space and thus breaks the whole environment.

That said, I am not necessarily against resetting the link on start-up.

yakutovicha:
> I would think that this is a very specific case in which one would hope that the user (developer) knows what they are doing. Here I am merely trying to prevent the current issue, that the aiidalab package was inadvertently installed into the user space and thus breaks the whole environment.
> 
> That said, I am not necessarily against resetting the link on start-up.

I am also not sure which is the best route to take here. But since there is no choice that is clearly better, I believe we should keep things as they are for the moment.

yakutovicha (approved):

PR #213, opened 2021-10-05
Do not make the apps/ folder Python importable.

csadorf:


yakutovicha (approved):
Thanks, @csadorf. I agree, that we should forbid any imports that include `apps` in it since the applications are now installed with standard python tools.

Somewhat in the same spirit, I created the following PR: https://github.com/aiidateam/aiida-prerequisites/pull/37. Please have a look

yakutovicha:
I believe it is what I wrote in the approval comment to https://github.com/aiidalab/aiidalab-docker-stack/pull/213. If that is not what you meant, then we should probably rediscuss.

PR #214, opened 2021-10-05
run.sh: add option to modify the default username inside the container.

yakutovicha:

PR #217, opened 2021-10-14
Use docker-compose files to manage local deployments.

csadorf:
The revised setup encourages a better separation of build and runtime
configuration.

The runtime configuration is configured via environment variables while
different build configurations (i.e. the development configuration) is
enabled by selecting the docker-compose.develop.yml file in addition to
the default docker-compose.yml file.

A user and app developer simply starts the deploy via: `./manage.py up`
while an AiiDAlab system developer can trigger the start of a deploy
with development versions installed via the `./manage.py --develop up`
command.  Note that this requires a different build of AiiDAlab, i.e.,
this command launches a different image.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -48,6 +48,13 @@ def _service_is_up(docker_compose, service):
 )
 @click.option("-p", "--project-name", help="Specify an alternative project name.")
 @click.option(
+    "--env-file",
+    type=click.Path(dir_okay=False, writable=True, path_type=Path),
+    default=".env",
+    help="The path of the env-file to use for configuration.",
+    show_default=True,
+)
+@click.option(
     "-v",
     "--verbose",
     count=True,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,3 +71,4 @@ Grant No. 676598.
 
 ![MARVEL](miscellaneous/logos/MARVEL.png)
 ![MaX](miscellaneous/logos/MaX.png)
+requirements-manage.txt
```
yakutovicha, line 74: I am not sure how did this line end up here.

docker-compose.yml
```diff
@@ -0,0 +1,16 @@
+version: "3.8"
+services:
+  aiidalab:
+    build:
+      context: .
+      args:
+        - aiidalab_home_version=v21.10.0
+    ports:
+      - ${AIIDALAB_PORT:-8888}:8888
+    environment:
```
yakutovicha, line 10: Can you also add the possibility to modify the default username? For an example, see #214

manage.py
```diff
@@ -0,0 +1,266 @@
...
+    )
+    click.echo(f"Written configuration to '{env_file}'.")
+
+
+@cli.command()
+@click.option(
+    "--restart", is_flag=True, help="Restart AiiDAlab in case that it is already up."
+)
+@click.pass_context
+def up(ctx, restart):
```
yakutovicha, line 159: Can you add an option to use a different `.env` file? This would allow starting multiple containers at the same time, I guess.

PR #227, opened 2021-11-23
check env file exist

unkcpz:


unkcpz:
Well, I think the more proper behaviour is there is an empty `.env` file in the repository.

csadorf (changes requested):
What issue is addressed by this PR? Generally I would not think that it should be a requirement that the `.env` file exists when first running the script, since the `configure` command is there to create the file.

unkcpz:
I guess sometimes we may want to just start the container with all default settings or the first time before running `configuration` I want to know what is being set default. Instead of a file exist check, maybe just put a default `.env` is better?

yakutovicha:
I assume due to the invention of [aiidalab-launch](https://github.com/aiidalab/aiidalab-launch) we can close this PR, right?

unkcpz:
@yakutovicha Yes, thanks!

PR #232, opened 2021-12-22
Correctly remove the content of .trash folder

yakutovicha:
Previously, the `find` was used to delete the trash folder's content.
However, `find` can't delete non-empty folders. A better approach
is to use `rm -rf /home/${SYSTEM_USER}/.trash/*` which should be
more robust.


yakutovicha:
since most of us are on vacation, I am going to proceed without a review. Pinging @csadorf and @unkcpz for the info.

csadorf:
Looks good, not sure what my original motivation to use `find` was.

Issue #225, opened 2021-11-15
AiiDAlab fails to start with error `File system loop detected; ‘/home/aiida/aiida’ is part of the same file system loop as ‘/home/aiida’.`


csadorf:
After I clean the `.trash/` and rerun the `manage.py` I got:

```
...
aiidalab_1  | + verdi daemon stop
aiidalab_1  | Profile: default
aiidalab_1  | Warning: Deleted apparently stale daemon PID file as its associated process<1276> does not exist anymore
aiidalab_1  | Waiting for the daemon to shut down... The daemon was not running.
aiidalab_1  | + verdi database migrate --force
aiidalab_1  | + verdi daemon start
aiidalab_1  | Starting the daemon... RUNNING
aiidalab_1  | *** Running /etc/my_init.d/80_prepare-aiidalab.sh...
aiidalab_1  | + export SHELL=/bin/bash
aiidalab_1  | + SHELL=/bin/bash
aiidalab_1  | + '[' '!' -e /home/aiida/.ipython/profile_default/startup/aiida_magic_register.py ']'
aiidalab_1  | + '[' '!' -e /home/aiida/apps ']'
aiidalab_1  | + '[' '!' -e /home/aiida/apps/home ']'
aiidalab_1  | + [[ -d /home/aiida/apps/home ]]
aiidalab_1  | + [[ ! -L /home/aiida/apps/home ]]
aiidalab_1  | + [[ '' == 1 ]]
aiidalab_1  | + reentry scan
aiidalab_1  | + '[' -e /home/aiida/.trash ']'
aiidalab_1  | + find -L /home/aiida -maxdepth 3 -name apps_meta.sqlite -delete
aiidalab_1  | find: File system loop detected; ‘/home/aiida/aiida’ is part of the same file system loop as ‘/home/aiida’.
aiidalab_1  | *** /etc/my_init.d/80_prepare-aiidalab.sh failed with status 1
aiidalab_1  | 
aiidalab_1  | *** Killing all processes...
aiidalab_1  | Nov 15 12:48:07 53e6cb9c8379 syslog-ng[115]: syslog-ng shutting down; version='3.13.2'
aiidalab-docker-stack_aiidalab_1 exited with code 1
```

_Originally posted by @unkcpz in https://github.com/aiidalab/aiidalab-docker-stack/issues/223#issuecomment-968879111_

csadorf:
[...]

PR #236, opened 2022-01-06
Update recommended approach for local deployment to use aiidalab-launch

csadorf:
- Replaces the documentation for launching AiiDALab locally to recommend the use of [aiidalab-launch](https://github.com/aiidalab/aiidalab-launch/) instead of manage.py
- Removes all scripts related to manage.py.


csadorf:
⚠️ @yakutovicha This approach will fail until we have made a new release that includes #225 . ⚠️ 

I would prefer to merge that first so that users do not get the impression that the launch tool is buggy from the start.

yakutovicha (approved):

PR #246, opened 2022-04-07
CI: Add pre-commit config.

csadorf:


csadorf:
Should fix CI failure which popped up on #245 .

yakutovicha (approved):
Looks good!

PR #257, opened 2022-06-21
Bump dependencies.

yakutovicha:
- Bump base image version to aiida-core:1.6.8
- Bump aiidalab package to 22.02.0
- Bump aiidalab-widgets-base to 1.3.3

WARNING: this changes the base image version to Ubuntu 20, which comes with a newer PostgreSQL. We do not yet have automated tools to migrate the PostgresSQL database.

PR #270, opened 2022-07-06
Roll-back to base image that does not require a PGSQL migration.

csadorf:
This commit partially reverts #257.


csadorf:
Depends on #271 .

csadorf:
After discussion with @yakutovicha this is likely not needed since the next release of the image should include the database migration.

PR #279, opened 2022-09-05
Implement full-stack image

csadorf:
Re-introduces a full-stack image (has all services included).


yakutovicha (changes requested):
Thanks, @csadorf. Please find below a couple of suggestions and comments.

Following our private discussion on the image name, I would propose to call it `base-with-services`. I think the name clearly delivers the information about the image content. A slightly modified version of your original proposition to name it `with-services`.

stack/full-stack/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+LABEL maintainer="AiiDAlab Team <aiidalab@materialscloud.org>"
+
+USER root
+WORKDIR /opt/
+
+ARG AIIDA_VERSION=2.0.0
+
+RUN mamba create -n aiida-core-services --yes \
+     aiida-core.services=${AIIDA_VERSION} \
+     rabbitmq-server=3.8.14 \
```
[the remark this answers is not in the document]
csadorf: It seems like this is not necessary (see https://github.com/aiidalab/aiidalab-docker-stack/pull/279/commits/02449081481cf050ef1e2c1b9bac7b5c6f7b0874 ).

stack/full-stack/before-notebook.d/30_start-rabbitmq.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
csadorf: It seems like this is not necessary (see https://github.com/aiidalab/aiidalab-docker-stack/pull/279/commits/02449081481cf050ef1e2c1b9bac7b5c6f7b0874 ).

dodo.py
```diff
@@ -28,6 +47,7 @@ def task_up():
                 "default": 8888,
                 "help": "Specify the AiiDAlab host port.",
             },
+            _REGISTRY_PARAM,
```
yakutovicha, line 50: I am probably nitpicking here, but if you decided to place registry param as a global variable, it probably makes sense to do the same for the port.

README.md
```diff
@@ -31,6 +31,12 @@ Finally, run the automated tests with `doit tests`.
 
 For manual testing, you can start the images with `doit up`, however please refer to the next section for a production-ready local deployment of AiiDAlab with aiidalab-launch.
 
+### Test on different architectures
+
+Images are built for multiple architectures (amd64 and aarch64) during continuous integration and pushed to the GitHub Container Registry (ghcr.io).
+To run automated tests or manual test against those images, simply specify the registry with the aformentioned `up` and `tests` commands (e.g. `doit up --registry=ghcr.io/).
```
yakutovicha, line 37: I think we should add the ability to specify the image tag/sha, when running `doit up`. I am not sure this is implemented in `dodo.py`.

README.md
```diff
@@ -31,6 +31,12 @@ Finally, run the automated tests with `doit tests`.
...
+Note: You may have to [log into the registry first](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry).
```
yakutovicha, line 38: Maybe just put the command here. Took me some time to find it in the link.

```suggestion
Note: You may have to log into the registry first (`docker login ghcr.io -u USERNAME`).
```

README.md
yakutovicha: Also, despite I am logged in, it didn't work for me:

```bash
$ docker login ghcr.io -u yakutovicha
Password:
Login Succeeded
$ doit up --registry=ghcr.io/
.  up
aiidalab Pulling
aiidalab Error
```

yakutovicha (commented):
Thanks, @csadorf, almost good to go. I have a few minor comments.

README.md
csadorf: Can you try to provide more information here? Maybe you can run the command manually to get more verbose output?
```console
REGISTRY=ghcr.io/ docker-compose up --detach
```

README.md
csadorf: I'd prefer not to, because the linked page provides more context, e.g., that one should obtain a PAT and with which access rights etc. I think by just showing a single command here – which may or may not work – we are oversimplifying this.

dodo.py
csadorf: The port is only used for the up task whereas the _REGISTRY_PRAM is used in all commands.

README.md
yakutovicha: ok, sounds good to me.

README.md
yakutovicha: Not much is shown.
```bash
REGISTRY=ghcr.io/ docker-compose up --detach
[+] Running 0/1
 ⠿ aiidalab Error                                                                                                                                                                                                                                                                                                                                                    0.8s
Error response from daemon: denied
```
It might be related to the token access rights. I will re-test this after lunch.

dodo.py
yakutovicha: ok, sounds good.

README.md
yakutovicha: Indeed, the issue was related to the token access rights.

The only issue that remains is this:


> I think we should add the ability to specify the image tag/sha, when running `doit up`. I am not sure this is implemented in `dodo.py`.

@csadorf can you please comment on this last point?

csadorf: This was implemented in https://github.com/aiidalab/aiidalab-docker-stack/pull/289 .

csadorf:
🧠 Reminder to myself: **"Rebase and merge"** !

yakutovicha (approved):
LGTM!

PR #281, opened 2022-09-05
Implement full-stack-lab image.

csadorf:
Combines the lab and full-stack images.

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -34,7 +34,7 @@ function "tags" {
 }
 
 group "default" {
-  targets = ["base", "base-with-services", "lab"]
+  targets = ["base", "base-with-services", "lab", "full-stack"]
 }
 
 target "base-meta" {
@@ -45,6 +45,10 @@ target "base-with-services-meta" {
 }
 target "lab-meta" {
   tags = tags("lab")
+}
+
+target "full-stack-meta" {
+  tags = tags("full-stack")
 }
 
 target "base" {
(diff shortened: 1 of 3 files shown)


docker-bake.hcl
```diff
@@ -33,7 +33,7 @@ function "tags" {
 }
 
 group "default" {
-  targets = ["base", "full-stack", "lab"]
+  targets = ["base", "full-stack", "lab", "full-stack-lab"]
```
yakutovicha, line 36: I would very much prefer to name **this** image `full-stack`

stack/full-stack-lab/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# syntax=docker/dockerfile:1
+FROM full-stack as base
+
+FROM  lab
+
+USER root
+
+COPY --from=base /opt/config-quick-setup.yaml /opt/
+COPY --from=base "${CONDA_DIR}/envs/aiida-core-services" "${CONDA_DIR}/envs/aiida-core-services"
```
yakutovicha, line 9: That can only work if we manage to install RMQ through conda for the arm64 architecture.

yakutovicha (commented):
More or less the same questions as for #279.

docker-bake.hcl
csadorf: Ok, unless you object I will adopt the following image names:

- base: only contains AiiDA
- base-with-services: like base + services (installed and started)
- lab: base + AiiDAlab components
- full-stack: AiiDAlab + services

stack/full-stack-lab/Dockerfile
csadorf: That appears to be the case. With https://github.com/aiidalab/aiidalab-docker-stack/pull/279/commits/02449081481cf050ef1e2c1b9bac7b5c6f7b0874 we build (albeit not necessarily test) all image for both architectures.

yakutovicha (approved):

PR #291, opened 2022-09-14
Loosen the awb version requirement.

### LINKED ISSUE #290: AWB not installable from the App store
Because the AWB version is defined as part of the pip constraint file as of #269, it is currently not possible to install AWB from the App store as of the today's release of the 1.4.0 version.

image

```sh
Defaulting to user installation because normal site-packages is not writeable
Processing /home/aiida/apps/aiidalab-widgets-base
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
ERROR: Cannot install aiidalab-widgets-base 1.4.0 (from /home/aiida/apps/aiidalab-widgets-base) because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested aiidalab-widgets-base 1.4.0 (from /home/aiida/apps/aiidalab-widgets-base)
    The user requested (constraint) aiidalab-widgets-base~=1.3.4

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

I suspect this also breaks the `aiidalab-launch` workflow as it tries to install AWB in the default configuration. Additionally, this prevents Apps to define minimum AWB version as a dependency.

TBH I don't understand why AWB is installed via pip into base image. Wouldn't it make more sense to install it as an AiiDAlab app into /home/aiida/apps during the initial AiiDAlab setup?

CC @yakutovicha @csadorf

yakutovicha:
fixes #290

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 aiidalab==22.7.1
+aiidalab-widgets-base~=1.3
 binaryornot~=0.4
 bokeh~=2.0
 bqplot~=0.12
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
 aiidalab==22.7.1
-aiidalab-widgets-base~=1.3.4
 binaryornot~=0.4
 bokeh~=2.0
 bqplot~=0.12


danielhollas:
I think a safer choice for a hotfix would be to loosen the version requirement. For example, I think the aiidalab-test-action depends on AWB being in the test image.

csadorf:
> I think a safer choice for a hotfix would be to loosen the version requirement. For example, I think the aiidalab-test-action depends on AWB being in the test image.

I agree.

csadorf (commented):
Like @danielhollas mentioned, I think it would be better to simply loosen the requirement to reduce the chances of unexpected side-effects.

danielhollas:
Just mentioning that another option would be to remove it from the requirements.txt file (which serves as pip constraints file) but install it automatically later in one of the setup scripts when launching the container. If it is not installed automatically, it would break apps that currently do not explicitly declare dependency on AWB in their setup.cfg (if such apps exists).

If the plan is to remove AWB from appstore then longer term all apps should declare it as a depency.

yakutovicha:
> > I think a safer choice for a hotfix would be to loosen the version requirement. For example, I think the aiidalab-test-action depends on AWB being in the test image.
> 
> I agree.

done

csadorf (approved):

danielhollas:
@yakutovicha hmm, but this change still won't allow me to install AWB 1.4.0 right? Should the constrain be `>=1.3,<2.0`?

csadorf:
> @yakutovicha hmm, but this change still won't allow me to install AWB 1.4.0 right? Should the constrain be `>=1.3,<2.0`?

The constraint `~=1.3` is essentially equivalent to `>=1.3,<2.0`.

danielhollas:
Ah, my bad, thanks!

Dne st 14. 9. 2022 11:51 dop. uživatel Carl Simon Adorf <
***@***.***> napsal:

yakutovicha:
The image is also released, it should be 22.8.1

danielhollas:
Thanks so much @yakutovicha for a speedy release. I've tested the new image and the problem is resolved.

unkcpz:
Do we need to loose more of this? I use the latest image and still have the issue to install the master branch of AWB. I got following error:

```
Defaulting to user installation because normal site-packages is not writeable
Obtaining file:///home/aiida/apps/aiidalab-widgets-base
  Preparing metadata (setup.py) ... done
ERROR: Cannot install aiidalab-widgets-base 1.4.0 (from /home/aiida/apps/aiidalab-widgets-base) because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested aiidalab-widgets-base 1.4.0 (from /home/aiida/apps/aiidalab-widgets-base)
    The user requested (constraint) aiidalab-widgets-base~=1.3.4

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
WARNING: You are using pip version 22.0.4; however, version 22.2.2 is available.
You should consider upgrading via the '/opt/conda/bin/python3.8 -m pip install --upgrade pip' command.
```

danielhollas:
Hmm, this fix was merged to the `support/aiida-1.x`. Seems to me it needs to be merged to master/develop as well? @yakutovicha

unkcpz:
I used image `aiidalab-docker-stack:22.8.0`, what is that build from, containing this change?

danielhollas:
As @yakutovicha writes above

> The image is also released, it should be 22.8.1

But tbh I am quite unsure of the current state of things and how things are versioned / released. (e.g. I don't see any commit that releases 22.8.1)

yakutovicha:
> But tbh I am quite unsure of the current state of things and how things are versioned / released. (e.g. I don't see any commit that releases 22.8.1)

The stack is now fully re-designed, please have a look at the readme: you can find new labels and tags. We keep support for the aiida-1.6.x in the `22.8.x` versions.

PR #295, opened 2022-09-14
Remove the VERSION variable from the build variables.

csadorf:
The version is determined by the tag for releases and by the state of the local git repository for local builds.


csadorf:
Should be merged after #296 .

csadorf:
Superseded by #296 .

PR #299, opened 2022-09-15
Fix: Permissions of the erlang cookie before start of the RMQ server.

csadorf:


csadorf:
@unkcpz This fixes the issue I encountered earlier on the Kubernetes deployment. Feel free to test it out yourself at https://aiida-tutorial-2022-test.aiidalab.xyz/ .

unkcpz:
@csadorf thanks! The change looks good to me. 
However, I lost the big picture of how RabbitMQ installed and configured in new stack image. From the last line, `mamba run -n aiida-core-services rabbitmq-server -detached`. It seems the rabbitmq is installed from conda forge into the env `aiida-core-services`, but it still lacks of the arm64 as I know (https://anaconda.org/conda-forge/rabbitmq-server)?

stack/base-with-services/before-notebook.d/30_start-rabbitmq.sh
```diff
@@ -6,6 +6,9 @@ RABBITMQ_DATA_DIR="/home/${NB_USER}/.rabbitmq"
 mkdir -p "${RABBITMQ_DATA_DIR}"
 fix-permissions "${RABBITMQ_DATA_DIR}"
```
unkcpz, line 7: Could you explain what this line does?

unkcpz (changes requested):

csadorf:
> @csadorf thanks! The change looks good to me. However, I lost the big picture of how RabbitMQ installed and configured in new stack image. From the last line, `mamba run -n aiida-core-services rabbitmq-server -detached`. It seems the rabbitmq is installed from conda forge into the env `aiida-core-services`, but it still lacks of the arm64 as I know (https://anaconda.org/conda-forge/rabbitmq-server)?

RabbitMQ is installed as part of the `base-with-services` image:
https://github.com/aiidalab/aiidalab-docker-stack/blob/79dfb38f97b4974e5525f8296409dc4ecc4a5a3b/stack/base-with-services/Dockerfile#L11-L16

The installation succeeds, because the docker image architecture is `linux/arm64`, which to my understanding can install from the `linux-64` conda-forge channel.

stack/base-with-services/before-notebook.d/30_start-rabbitmq.sh
csadorf: The [fix-permissions script](https://github.com/jupyter/docker-stacks/blob/main/base-notebook/fix-permissions) ensures that the permissions of the data dir are set such that the user's group has read and write access to it.

unkcpz (commented):
For the change itself, looks all good. But still paranoid about the rabbitmq installed from conda for arm64, we are discussing it offline.

unkcpz (approved):

PR #303, opened 2022-09-16
Fix: Use 'core.direct' instead of 'direct' to setup scheduler.

csadorf:


unkcpz (approved):
Looks all good to me, cheers.

PR #307, opened 2022-09-21
Test commit with minor change to open PR

mbercx:


csadorf:
The permission error comes from the container registry. This happens when there is a second CI run that tries to push at the same time. It should work just fine if you restart the workflow as long as there is not a second one running in parallel.

mbercx:
Closing this PR since it's opened from my fork, which means the CI won't have the required secrets for the dockerhub deployment. Opened #308 instead.

csadorf:
Correct, I didn't realize the PR was coming from your fork.

PR #309, opened 2022-09-22
Fix: Set correct home directory for `localhost` `workdir`

mbercx:


csadorf (approved):
Approved pending tests pass.

PR #311, opened 2022-09-27
Remove AiiDA version defaults in Dockerfiles

mbercx:
Since the AiiDA `core` version is obtained from the `build.json`,
these are just defaults that are rarely used. So at this point
the default are only a source of confusion and over-specification
and it is better to remove them.


csadorf (approved):
Thanks a lot!

PR #315, opened 2022-10-07
autocomplete set using bash_source

### LINKED ISSUE #314: The autocomplete set in `base` image  is not working
We have this issue both on Azure deployment and with the container launched by `aiidalab-launch`.
It seems it is configured here https://github.com/aiidalab/aiidalab-docker-stack/blob/d06ea7c9e6b660f2774224ea494a836f54fcfc3b/stack/base/Dockerfile#L29, but not working.

unkcpz:
fixes #314

### DIFF
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -26,7 +26,7 @@ ENV PIP_CONSTRAINT=/opt/requirements.txt
 
 # Enable verdi autocompletion.
 RUN mkdir -p "${CONDA_DIR}/etc/conda/activate.d" && \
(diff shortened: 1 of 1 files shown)


danielhollas:
@unkcpz Would you mind explaining the fix? Why doesn't the `source` work in this case?

unkcpz:
Hi @danielhollas, I don't know why exactly sorry, only know it is a setting from https://click.palletsprojects.com/en/8.1.x/shell-completion/

mbercx (approved):
Thanks @unkcpz! LGMT 🚀

PR #316, opened 2022-10-12
aiida-core installed by pip and bump python version 3.10

### LINKED ISSUE #265: Support Python 3.11
Adopt Python 3.11 as the primary Python version in AiiDAlab.

## TODO list

- [x] Bump watchdog dependency in `aiidalab` package https://github.com/aiidalab/aiidalab/pull/328
- [x] Verify that `optimade-client` works with Python 3.10 https://github.com/CasperWA/voila-optimade-client/pull/474
- [x] Verify AWB and QeApp can be installed with Python 3.11, this can be done by automatic integration test on this repo
- [ ] Manually test that everything works
- [ ] Bump Python version in Docker stack #455

unkcpz:
fixes #265 

I am more kin to using pip to install `aiida-core` which is more well-tested and has fewer restrictions when comes to conda dependencies (for example if we want to have `arm64` supported). 
The current blocker of using python3.10 is that the wrapt 1.11 does not have python3.10 support which addressed by https://github.com/aiidateam/aiida-core/pull/5698. It also shows that conda install of aiida-core is not well tested.

### DIFF
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -1,4 +1,4 @@
-ARG BASE=jupyter/minimal-notebook:python-3.9.4
+ARG BASE=jupyter/minimal-notebook:python-3.10.6
 FROM ${BASE}
 
 LABEL maintainer="AiiDAlab Team <aiidalab@materialscloud.org>"
@@ -10,13 +10,10 @@ ARG AIIDA_VERSION
 
 # Install the shared requirements.
 COPY requirements.txt .
-RUN mamba install --yes \
-     aiida-core=${AIIDA_VERSION} \
-     --file requirements.txt \
-     && mamba clean --all -f -y && \
+RUN pip install --no-cache-dir aiida-core==${AIIDA_VERSION} \
+     -r requirements.txt && \
      fix-permissions "${CONDA_DIR}" && \
      fix-permissions "/home/${NB_USER}"
-
 
 # Pin shared requirements in the base environemnt.
 RUN cat requirements.txt | xargs -I{} conda config --system --add pinned_packages {}
--- a/build.json
+++ b/build.json
@@ -1,7 +1,7 @@
 {
   "variable": {
     "PYTHON_VERSION": {
-      "default": "3.9.4"
+      "default": "3.10.6"
     },
     "AIIDA_VERSION": {
       "default": "2.0.3"
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -21,7 +21,7 @@ variable "REGISTRY" {
 }
 
 variable "PLATFORMS" {
-  default = ["linux/amd64", "linux/arm64"]
+  default = ["linux/amd64"]
 }
 
 function "tags" {
--- a/stack/full-stack/Dockerfile
+++ b/stack/full-stack/Dockerfile
@@ -10,7 +10,7 @@ COPY --from=base "${CONDA_DIR}/envs/aiida-core-services" "${CONDA_DIR}/envs/aiid
 COPY --from=base /usr/local/bin/before-notebook.d /usr/local/bin/before-notebook.d
 
 RUN fix-permissions "${CONDA_DIR}"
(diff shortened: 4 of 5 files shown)


stack/base/Dockerfile
```diff
@@ -10,14 +10,11 @@ ARG AIIDA_VERSION
 
 # Install the shared requirements.
 COPY requirements.txt .
-RUN mamba install --yes \
-     aiida-core=${AIIDA_VERSION} \
-     --file requirements.txt \
-     && mamba clean --all -f -y && \
+RUN pip install --no-cache-dir aiida-core==${AIIDA_VERSION} \
+     -r requirements.txt && \
```
danielhollas, line 14: This also means that things in `requirement.txt` will be installed by pip. Is that what you wanted?

unkcpz: You are right. To be honest, I have no idea why this `requirements.txt` is needed, it has `aiida-core` and `pip` included, but aiida-core is already installed with specifying the version.

danielhollas: It is used below as a contraint file for conda and pip. But it is indeed strange that aiida-core version is specified in two places. This has already been noted in #311. Perhaps `requirements.txt` needs to be generated dynamically in the Dockerfile?

cc @csadorf

unkcpz:
There is a big change here. The new base `jupyter/minimal-notebook` start to support `linux/arm64/v8` which is not for the old base image. This is also the wired reason that the arm64 build pass and what is being downloaded in my arm64 machine is actually amd64. 
I propose to remove the support to `arm64` and add it back after I also start and test by using the self-hosted CI runner.

unkcpz:
I'll revert this back using only mamba, I think the dependent issue needs to be fixed from aiida-core side.

danielhollas:
@unkcpz I guess I don't see a reason why not use pip to install AiiDA, if that is a better supported installation method, and I guess it makes sense given that dependencies of AiiDAlab apps are installed via pip. I wonder if that would on the other hand solve the issue that I reported in #320 with pymatgen. 

I do worry a bit about the unintended consequences of this, as pip and conda sometimes interact strongly, but I don't have a deep experience with this. Definitely needs testing.

stack/base/Dockerfile
csadorf: Indeed, the `requirements.txt` file serves a double function as constraints file. The aiida-core version is not pinned using this file, but only constraint, as you can see here: https://github.com/aiidalab/aiidalab-docker-stack/blob/a70d72b4b4be9d4d70ee842d6ad741d5bf319e7f/stack/base/requirements.txt#L1
It would need to be updated if and only if the major version is changed.

The only other requirement in that file is the pinned pip version.

csadorf:
> @unkcpz I guess I don't see a reason why not use pip to install AiiDA, if that is a better supported installation method, and I guess it makes sense given that dependencies of AiiDAlab apps are installed via pip. I wonder if that would on the other hand solve the issue that I reported in #320 with pymatgen.
> 
> I do worry a bit about the unintended consequences of this, as pip and conda sometimes interact strongly, but I don't have a deep experience with this. Definitely needs testing.

Installing aiida-core with pip means that all of its (long list of) dependencies are also installed via pip which means that the conda environment is basically converted into a pip-managed environment rendering it almost unusable for further conda installations.

unkcpz:
> Installing aiida-core with pip means that all of its (long list of) dependencies are also installed via pip which means that the conda environment is basically converted into a pip-managed environment rendering it almost unusable for further conda installations.

I see, thanks! However, I still see no advantages of using conda over pip in installing `aiida-core`.
Since we encounter some issues with installing aiida-core by conda, and pip-manager as I know is more widely used in AiiDA community. I think we can stick to the strategy to use conda/mamba for the non-python tools such as postgreSQL, rabbitmq, QE and for creating virtual environments. Meanwhile, we stick to pip to install the python packages and AiiDAlab apps.

danielhollas:
Superseded by #337

PR #318, opened 2022-10-13
Using base image with python version 3.9.13

unkcpz:
As mentioned https://github.com/aiidalab/aiidalab-docker-stack/pull/316#issuecomment-1277911476,
The new base jupyter/minimal-notebook start to support linux/arm64/v8 which is not for the old base image. This is also the wired reason that the arm64 build pass and what is being downloaded in my arm64 machine is actually amd64.
I propose to remove the support to arm64 and add it back after I also start and test by using the self-hosted CI runner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ For manual testing, you can start the images with `doit up`, however we recommen
 
 ### Continuous integration
 
-Images are built for `linux/amd64` and `linux/arm64` during continuous integration for all pull requests into the default branch and pushed to the GitHub Container Registry (ghcr.io) with tags `ghcr.io/aiidalab/*:pr-###`.
+Images are built for `linux/amd64` during continuous integration for all pull requests into the default branch and pushed to the GitHub Container Registry (ghcr.io) with tags `ghcr.io/aiidalab/*:pr-###`.
 You can run automated or manual tests against those images by specifying the registry and version for both the `up` and `tests` commands, example: `doit up --registry=ghcr.io/ --version=pr-123`.
 Note: You may have to [log into the registry first](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry).
 
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -21,7 +21,7 @@ variable "REGISTRY" {
 }
 
 variable "PLATFORMS" {
-  default = ["linux/amd64", "linux/arm64"]
+  default = ["linux/amd64"]
 }
 
 function "tags" {
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -1,4 +1,4 @@
-ARG BASE=jupyter/minimal-notebook:python-3.9.4
+ARG BASE=jupyter/minimal-notebook:python-3.9.13
 FROM ${BASE}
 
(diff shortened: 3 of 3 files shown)


danielhollas:
> The new base jupyter/minimal-notebook start to support linux/arm64/v8 which is not for the old base image. 

I don't think I am qualified to review this, but in any case I am quite confused by this sentence.

unkcpz:
> I don't think I am qualified to review this

Sure you are, this is an open source project, and you have exceeded qualification to review this from my perspective. :)

> > The new base jupyter/minimal-notebook start to support linux/arm64/v8 which is not for the old base image.
>
> but in any case I am quite confused by this sentence.

My bad. I mean `jupyter/minimal-notebook:python-3.9.4` which is used by `base` image only has `amd64` [1].
In this PR I change it to `jupyter/minimal-notebook:python-3.9.13` has both `linux/amd64` and `liunx/arm64/v8` supported.[2]

[1] https://hub.docker.com/r/jupyter/minimal-notebook/tags?page=1&name=3.9.4
[2] https://hub.docker.com/r/jupyter/minimal-notebook/tags?page=1&name=3.9.13

danielhollas (commented):
Got it, thanks. Got confused since you're removing the `linux/arm64` support while at the same time bumping to the image that supports it. :-)

> Sure you are, this is an open source project, and you have exceeded qualification to review this from my perspective. :)
:blush: 

Leaving approve for others to weigh in whether support should be removed here.

unkcpz:
> Leaving approve for others to weigh in whether support should be removed here.

I'd say it has to be removed otherwise the build will fail in any case. The original support is a "fake" support, the docker [buildx will fall to use the one with wrong architecture](https://github.com/docker/buildx/issues/844) and set the image digest to wrong image.

yakutovicha:
> Got it, thanks. Got confused since you're removing the `linux/arm64` support while at the same time bumping to the image that supports it. :-)

I am also a bit confused here. Why does switching to the base image that supports the `arm` architecture require switching off the `arm` build?

unkcpz:
> I am also a bit confused here. Why does switching to the base image that supports the arm architecture require switching off the arm build?

If the image `jupyter/minimal-notebook` not support arm64, when building for arm64 the amd64 will be used and all build passes.
When there is the image support arm64, it will use the correct one but our stack doesn't support arm64 yet such as the rabbitmq don't have conda forge for arm64.

danielhollas (approved):
>  When there is the image support arm64, it will use the correct one but our stack doesn't support arm64 yet such as the rabbitmq don't have conda forge for arm64.

So if I understand correctly, arm64 support will be added back in some future PR and besides upgrading the base image (done here) is blocked on other dependencies (RabbitMQ?).

Approving under the assumption that my understanding is correct. :-)

unkcpz:
> So if I understand correctly, arm64 support will be added back in some future PR and besides upgrading the base image (done here) is blocked on other dependencies (RabbitMQ?).

Thanks, @danielhollas. Yes, this is what my plan is. But for rabbitmq, I don't think it will have conda forge package supported for arm64 soon, even if it has, it may not support `<3.8.15` as we needed for AiiDA. Probably the best way to install rabbitmq is to roll back to using `apt` with the latest version and change the timeout setting in the configuration.

yakutovicha (approved):
LGTM, thanks @unkcpz!

PR #319, opened 2022-10-14
Slim image based on mambaorg/micromamba

unkcpz:
This is an image similar to `base` but based on `mambaorg/micromamba` which is lightweight than `jupyter/minimal-notebook` without jupyter things pre-installed. 
I think it is suitable as the basic "aiida only" image, has the advantages of:
1. The system user can be changed, and set as `aiida` as default.
2. Size is slim compared with `base` (741MB v.s 1.9 GB)
3. Easy to control the startup scripts.

I also give a `docker-compose` file and tested it in my workstation so that it can work well with the Postgre container and rabbitmq container in a composed stack.

- [ ] docker-compose image name configurable.
- [ ] Duplicate AiiDA config files with `base` image.
- [ ] Tests added


danielhollas:
@unkcpz do you know why the `base` image is so large compared to the `slim` one? Is it all the Jupyter stuff? If not perhaps we could trim it down, but still be based on the jupyter/minimal-notebook?

unkcpz:
We have the docker image for `aiida-core`, which are like the `base` and `base-with-services` from this repo. 
We decided still keep using the jupyter notebook docker stack as the base image which relieves the burden to check the compatibility of jupyter components (e.g. `jupyter-client`, `jupyter-server`).

unkcpz:
One more change we can have for this repo is we don't release the `base` and `base-with-services` to dockerhub and ghcr registry anymore. They only used internally for build `lab` and `full-stack`.

PR #322, opened 2022-11-10
Bumper aiida-core version to v2.1.2

### LINKED ISSUE #321: support aiida-core 2.1
aiida-core 2.1 is released with many exciting features, we need to support it ASAP.

unkcpz:
fixes #321

### DIFF
--- a/docker-compose.lab.yml
+++ b/docker-compose.lab.yml
@@ -11,7 +11,7 @@ services:
         volumes:
             - aiida-postgres-db:/var/lib/postgresql/data
         healthcheck:
-            test: ["CMD-SHELL", "pg_isready"]
+            test: ["CMD-SHELL", "pg_isready -U $${POSTGRES_USER}"]
             interval: 5s
             timeout: 5s
             retries: 10
--- a/stack/base/config-quick-setup.yaml
+++ b/stack/base/config-quick-setup.yaml
@@ -1,6 +1,6 @@
 ---
 db_engine: postgresql_psycopg2
-db_backend: psql_dos
+db_backend: core.psql_dos
 db_host: database
 db_port: 5432
 su_db_username: pguser
--- a/build.json
+++ b/build.json
@@ -4,7 +4,7 @@
       "default": "3.9.4"
     },
     "AIIDA_VERSION": {
-      "default": "2.1.0"
+      "default": "2.1.2"
     }
   }
 }
--- a/build.json
+++ b/build.json
@@ -4,7 +4,7 @@
       "default": "3.9.4"
     },
     "AIIDA_VERSION": {
-      "default": "2.0.4"
+      "default": "2.1.0"
     }
   }
 }


yakutovicha (commented):
@unkcpz, could you have a look at the test's failure, please?

unkcpz:
Don't know what take this so long, all fine to run on my workstation.

danielhollas:
From the test failure:

>  ../../../micromamba-root/envs/aiidalab-docker-stack/lib/python3.10/site-packages/pytest_docker/plugin.py:24: Exception

That's strange, I thought we're still on Python 3.9, why does this path lead to Python 3.10?

Just a thought, it would be great if we could somehow automatically run the AWB tests here to make sure the new aiida-core versions do not somehow break the widgets.

unkcpz:
> That's strange, I thought we're still on Python 3.9, why does this path lead to Python 3.10?

Nice catch, probably because I trigger the workflow by hand? And also there is a `catch-env` of mamba installing for test CI workflow. Still don't know why it is happened, but will play around a bit.

unkcpz:
I fixed the failed test, it was a api change of `core.psql`. But this should not merge since there is a bug fixed and 2.1.2 released hopefully today which is critical to the QE work chain.

danielhollas (approved):
Thanks @unkcpz.

@yakutovicha just thinking about a proper workflow for updating these Docker images, are we able to somehow run the AWB tests on this branch?

> But this should not merge since there is a bug fixed and 2.1.2 released hopefully today which is critical to the QE work chain.

Would be good to change the PR to draft just to make it obvious.

docker-compose.lab.yml
```diff
@@ -11,7 +11,7 @@ services:
         volumes:
             - aiida-postgres-db:/var/lib/postgresql/data
         healthcheck:
-            test: ["CMD-SHELL", "pg_isready"]
+            test: ["CMD-SHELL", "pg_isready -U $${POSTGRES_USER}"]
```
danielhollas, line 14: Why was this not needed before?

stack/base/config-quick-setup.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 db_engine: postgresql_psycopg2
-db_backend: psql_dos
+db_backend: core.psql_dos
```
danielhollas, line 3: Why did this work before? Or is this change to just squash a deprecation warning?

danielhollas (approved):
Couple non-blocking questions. Not sure why the CI did not run.

unkcpz:
The CI test actually run https://github.com/aiidalab/aiidalab-docker-stack/actions/runs/3487444767, but very wired it is not updated here.

docker-compose.lab.yml
unkcpz: Because this is not a critical issue, without this it will pop a warning as discussed https://github.com/peter-evans/docker-compose-healthcheck/issues/16, the healthcheck still pass somehow.

stack/base/config-quick-setup.yaml
unkcpz: It was not working before, this is the reason why simply bump the version to 2.1 fail the test at first place. This is a API change introduced in aiida-core 2.1 https://github.com/aiidateam/aiida-core/blob/main/CHANGELOG.md#pluginable-data-storage-backends

danielhollas: I am very surprised they did not change the entry point in a backwards compatible manner, as with previous analogous changes. Is that actually a bug that should be reported? cc @sphuber 

(the change here is fine of course, thank you for the explanation)

sphuber: Hmm, this we might have missed. The change was intended to be backwards-compatible. Existing profiles with `psql_dos` *are* automatically migrated to `core.psql_dos`, but indeed, the option for `verdi quicksetup/setup` now no longer accepts `psql_dos`.

danielhollas: @sphuber thanks for chiming in. Note that while migrating we also hit some other backwards incompatible changes when setting up the computer, I think transport `core.local` and scheduler, e.g. `core.slurm`, see changes in this PR.

https://github.com/ispg-group/aiidalab-ispg-docker-stack/pull/11/files#diff-602080df8884ed8d19826da3ccee7524f675094c9909ddcb3b9ebe8b6847696aR13

sphuber: Yeah, same problem. It is quite tricky to make these things backwards compatible. For option values we could potentially implement some deprecation pathway by temporarily allowing old values and in the command body detecting them and converting to the new entry point string. However, there are also subcommands that are based on the entry point string, for example `verdi computer configure local` now has become `verdi computer configure core.local` and this is very tricky to keep a deprecation pathway.

But in the end, all of this was added in v2.0 (https://github.com/aiidateam/aiida-core/commit/16daad08e63f0325ecd3fdaed7b2f7f8001ddaa7) so we considered these small breaking changes to be acceptable in a major version bump.

unkcpz:
@danielhollas thanks for the review and discussion.

PR #324, opened 2022-11-22
Fix: Add missing JUPYTER_CONFIG_PATH

danielhollas:
See this comment for details https://github.com/aiidalab/aiidalab/issues/322#issuecomment-1323417290

This perhaps should be fixed in the `jupyter/minimal-notebook` image itself, but in the meantime we fix it here.

### DIFF
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -52,6 +52,10 @@ if [ -f /opt/bin/load-singlesshagent.sh ]; then\n\
    . /opt/bin/load-singlesshagent.sh\n\
 fi\n' >> "/home/${NB_USER}/.bashrc"
 
+# Workaround for missing Jupyter config path,
+# needed for locally installed notebook extensions (e.g. nglview)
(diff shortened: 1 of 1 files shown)


danielhollas:
I am not sure why the build failed, but it might be because it is building from my fork and lacking some permissions.

@unkcpz would you mind giving me write access to this repo so that I can push my branch directly here?

unkcpz:
> @unkcpz would you mind giving me write access to this repo so that I can push my branch directly here?

Sure, yes. And I think you are correct about the build failing, it seems the token is not accessible from the forked repo. But I think it must be a more elegant way to tackle it, that makes no sense that only build from this repo can successfully test.

danielhollas:
@unkcpz thanks for adding me! I am closing this PR in favour of #325 

> But I think it must be a more elegant way to tackle it, that makes no sense that only build from this repo can successfully test.

This might be tricky, since the credentials are needed to push the built images to the `ghcr.io`. I think it is fine like this for now.

Issue #323, opened 2022-11-14
profiles remove does not delete the volume


superstar54:
## How to reproduce
- created a profile `aiida2.0.3`
- start, `aiidalab-launch start -p aiida2.0.3`
- stop and remove the container, `aiidalab-launch stop -r aiida2.0.3`
- remove the profile, `aiidalab-launch profiles remove aiida2.0.3`
- list the docker volume, `docker volume list`

## Expected behavior
The volume still exists:
```bash
local     aiidalab_aiida2.0.3_conda
local     aiidalab_aiida2.0.3_home
```

superstar54:
@danielhollas Thanks for the fast response. `reset` works, thanks!

> For aiidalab-launch related issues, please open them on the aiidalab-launch repository.

My mistake to open this issue here.

PR #325, opened 2022-11-23
Fix: Add missing path for $HOME/.local/bin

danielhollas:
See this comment for details https://github.com/aiidalab/aiidalab/issues/322#issuecomment-1323417290

https://docs.jupyter.org/en/latest/use/jupyter-directories.html

This perhaps should be fixed in the jupyter/minimal-notebook image itself, but in the meantime we fix it here.

This PR supersedes #324, that was open from my fork and tests were not passing.

### DIFF
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -53,7 +53,7 @@ if [ -f /opt/bin/load-singlesshagent.sh ]; then\n\
 fi\n' >> "/home/${NB_USER}/.bashrc"
 
 # Add ~/.local/bin to PATH where the dependencies get installed via pip
-RUN echo "export PATH\${PATH}:/home/${NB_USER}/.local/bin"
+RUN echo "export PATH=\${PATH}:/home/${NB_USER}/.local/bin" >> "/home/${NB_USER}/.bashrc"
 
 # Workaround for missing Jupyter config path,
 # needed for locally installed notebook extensions (e.g. nglview)
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -53,7 +53,7 @@ if [ -f /opt/bin/load-singlesshagent.sh ]; then\n\
 fi\n' >> "/home/${NB_USER}/.bashrc"
 
 # Add ~/.local/bin to PATH where the dependencies get installed via pip
-RUN echo "export PATH=\${PATH}:/home/${NB_USER}/.local/bin" >> "/home/${NB_USER}/.bashrc"
+ENV PATH=${PATH}:/home/${NB_USER}/.local/bin
 
 # Workaround for missing Jupyter config path,
 # needed for locally installed notebook extensions (e.g. nglview)
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -52,6 +52,10 @@ if [ -f /opt/bin/load-singlesshagent.sh ]; then\n\
    . /opt/bin/load-singlesshagent.sh\n\
 fi\n' >> "/home/${NB_USER}/.bashrc"
 
+# Workaround for missing Jupyter config path,
+# needed for locally installed notebook extensions (e.g. nglview)
+ENV JUPYTER_CONFIG_PATH=/home/jovyan/.local/etc/jupyter
+
 USER ${NB_USER}
 
 WORKDIR "/home/${NB_USER}"
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -53,7 +53,7 @@ if [ -f /opt/bin/load-singlesshagent.sh ]; then\n\
 fi\n' >> "/home/${NB_USER}/.bashrc"
 
 # Add ~/.local/bin to PATH where the dependencies get installed via pip
-RUN echo "export PATH=/home/${NB_USER}/.local/bin:\${PATH}"
+RUN echo "export PATH\${PATH}:/home/${NB_USER}/.local/bin"
 
 # Workaround for missing Jupyter config path,
 # needed for locally installed notebook extensions (e.g. nglview)
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -57,7 +57,7 @@ RUN echo "export PATH=\${PATH}:/home/${NB_USER}/.local/bin" >> "/home/${NB_USER}
 
 # Workaround for missing Jupyter config path,
 # needed for locally installed notebook extensions (e.g. nglview)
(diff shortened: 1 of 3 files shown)


danielhollas:
@unkcpz it seems my current permissions are still not enough to publish the built image to GitHub packages.

Would you mind cherry picking my commit and making a new PR yourself?

unkcpz:
> @unkcpz it seems my current permissions are still not enough to publish the built image to GitHub packages.

You mean from your forked repo? I think it is not related to the permission. The reason is that the token for upload to ghcr is set in this repo. 
If you want to have the pull_request test pass, you may need to get your PAT, please check https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry see if that help.

danielhollas:
@unkcpz this PR is from the branch on this repo, so there should be no issues with permissions in principle. I do not know why the test still doesn't pass. If you could open a new PR from your branch on this repo we could verify that.

Interestingly, the test pass in the other PR that I opened today.

unkcpz:
> Interestingly, the test pass in the other PR that I opened today.

Sorry, which one? I think PRs you push from this repo is all pass, but still not pass the one from your forked repo, correct?

danielhollas:
> Sorry, which one? I think PRs you push from this repo is all pass, but still not pass the one from your forked repo, correct?

Sorry, It seems the tests are indeed passsing here. I think the Github UI is very confused, because if you go to the [checks tab](https://github.com/aiidalab/aiidalab-docker-stack/pull/325/checks), it still shows the failing tests. (I think it might be because I opened two PRs, this one and the old one #324 from the fork, but the commit hashes are the same in both PRs so Github is confused).

In any case, this is good to go. Could you test this together with the fix on AiiDAlab? 
https://github.com/aiidalab/aiidalab/pull/323

unkcpz:
> In any case, this is good to go. Could you test this together with the fix on AiiDAlab?
https://github.com/aiidalab/aiidalab/pull/323

I can confirm that the issue exists on the production JupyterHub deployment, once I close the server and start a new one the dependencies are all gone. I'll get to #323 and test it.

unkcpz:
Hi @danielhollas, I think the path added can be tested with the pytest implemented here. If you don't mind I'll add a commit to this PR.

danielhollas:
@unkcpz go for it, thanks!

stack/base/Dockerfile
```diff
@@ -52,6 +52,13 @@ if [ -f /opt/bin/load-singlesshagent.sh ]; then\n\
    . /opt/bin/load-singlesshagent.sh\n\
 fi\n' >> "/home/${NB_USER}/.bashrc"
 
+# Add ~/.local/bin to PATH where the dependencies get installed via pip
+RUN echo "export PATH=\${PATH}:/home/${NB_USER}/.local/bin" >> "/home/${NB_USER}/.bashrc"
+
+# Workaround for missing Jupyter config path,
+# needed for locally installed notebook extensions (e.g. nglview)
+ENV JUPYTER_CONFIG_PATH=/home/${NB_USER}/.local/etc/jupyter
```
unkcpz, line 60: Hi @danielhollas, this will override original `JUPYTER_CONFIG_PATH` if it was set? This is a search path so can also be like `$PATH` and set as `${JUPYTER_CONFIG_PATH}:/home/${NB_USER}/.local/etc/jupyter`? Can you test it?

unkcpz:
Tests are failed, seems the NB_USER set to system user somewhere, I'll check. The Jupyter path seems also not set successfully, @danielhollas can you check at this?

danielhollas:
I'll take a look.

stack/base/Dockerfile
danielhollas: I don't think this path is set, but I agree it's better to do it as you propose, thanks.

danielhollas:
@unkcpz I have fixed the new tests, there were issues with escaping. Also, the `.bashrc` file is only sourced for interactive session, so I have instead set the PATH in Dockerfile via ENV. The tests are passing now, could you test on real deployment?

unkcpz:
> I have fixed the new tests, there were issues with escaping. 

@danielhollas thanks! Yes, clever fixes, I didn't notice. Sorry for the late reply, I was on the train all day, and it was not comfortable working on the train. 🤣

> The tests are passing now, could you test on real deployment?

I'd rather not test on real deployment, but will for sure test it locally launched using aiidalab-launch with installing QeApp, for the issue solved here it should be the same.

unkcpz:
Hi @danielhollas, it is wierd that when I start the `full-stack` image build from this branch, it start a jupyter lab rather than the notebook mode. I think changing the Jupyter configure path may the reason cause it?

unkcpz:
I think the code https://github.com/aiidalab/aiidalab-docker-stack/blob/4556c625571d3a78815d1f44b37107f2b991d897/stack/lab/Dockerfile#L77-L79 go after the new jupyter path you set in this PR.

danielhollas:
Interesting. Feel free to push to this PR, I won't get to it till tomorrow.

danielhollas:
>  real deployment

Right, sorry I wasn't clear, I just meant testing locally with the QeApp :-)

unkcpz:
Oh, no! I think the issue exist with the main branch. The image in dockerhub already has the issue.

danielhollas:
> the JUPYTER_CONFIG_PATH has higher priority if it is set

It's a bit weird because this variable should only add additional paths, so unless there is a config file, it shouldn't overwrite the default? I think the easiest fix is to modify to lab Dockerfile to create the config in the JUPITER_CONFIG_PATH location.

unkcpz:
>  I think the easiest fix is to modify to lab Dockerfile to create the config in the JUPITER_CONFIG_PATH location.

I open an issue for the notebook mode issue. But I think here you are correct, the setting for Jupyter should go to `lab` stack.

unkcpz:
With the new jupyterhub backend, the change for jupyter_path_config is not necessary as mentioned in https://docs.jupyter.org/en/latest/use/jupyter-directories.html#envvar-JUPYTER_PATH
The `JUPYTER_PATH` will be `JUPYTER_DATA_DIR` or (if not set) ~/.local/share/jupyter/ (respects $XDG_DATA_HOME). It is the path where the JS is installed and loaded.

danielhollas:
Cool. Please verify that by running `jupyter --path`

unkcpz:
Yep! The path is there.
image

unkcpz (commented):
Hi @danielhollas, I think this one is ready, correct? can you review it since the initial jupyter_path part is removed?

unkcpz (approved):

PR #326, opened 2022-11-23
Fix Python version in build.json

danielhollas:
This was an omission in #318. The Python version specified in `build.json` is used to automatically generate the image tags on Dockerhub. So currently, for example the image `aiidalab/full-stack:python-9.4` actually corresponds to python-9.13

https://hub.docker.com/r/aiidalab/full-stack/tags

image

### DIFF
--- a/build.json
+++ b/build.json
@@ -1,7 +1,7 @@
 {
   "variable": {
     "PYTHON_VERSION": {
-      "default": "3.9.4"
+      "default": "3.9.13"
     },
     "AIIDA_VERSION": {
       "default": "2.1.2"


unkcpz (changes requested):
I think the default `BASE` in `stack/base/Dockerfile` is set in `docker-bake.hcl` for bulid the image. Remove the `ARG BASE=jupyter/minimal-notebook:python-3.9.13` should be fine and good to have only one place to set the version.

unkcpz:
If seems the variable can not pass directly to `FROM`, but need to `ARG` first, https://stackoverflow.com/questions/36556766/can-a-variable-be-used-in-docker-from this may help.

unkcpz (approved):
Looks all good! Thanks!

danielhollas:
@unkcpz thanks! I just confirmed that the images published before this PR were still on Python 9.4, so the PR in #318 didn't work.
The fix with `ARG BASE` was the same as what Marnik did in #311 for `AIIDA_VERSION`.

I'll open a new PR where I will do the same fix for `AIIDALAB_VERSION` and `AIIDALAB_HOME_VERSION` in `stack/lab/Dockerfile`.

PR #327, opened 2022-11-24
Move AIIDALAB_VERSION and AIIDALAB_HOME_VERSION to build.json

danielhollas:
Also removed their defaults in Dockerfile, same fix as done in #311 for AIIDA_VERSION to reduce duplication.

### DIFF
--- a/build.json
+++ b/build.json
@@ -10,7 +10,7 @@
       "default": "22.08.0"
     },
     "AIIDALAB_HOME_VERSION": {
-      "default": "22.08.0"
+      "default": "22.11.0"
     }
   }
 }
--- a/build.json
+++ b/build.json
@@ -5,6 +5,12 @@
     },
     "AIIDA_VERSION": {
       "default": "2.1.2"
+    },
+    "AIIDALAB_VERSION": {
+      "default": "22.08.0"
+    },
+    "AIIDALAB_HOME_VERSION": {
+      "default": "v22.08.0"
     }
   }
 }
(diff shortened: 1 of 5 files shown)


unkcpz:
As I proposed at https://github.com/aiidalab/aiidalab-docker-stack/issues/332, the test can be simplified without checking the `variant` parameter.

danielhollas:
Hi @unkcpz,

this one is ready to go as well. I have bumped the aiidalab-home version and the version check now passes.

> Hi @danielhollas, as I proposed at #332, the test can be simplified without checking the `variant` parameter.

I have kept the tests as is for now, I think we need more discussion about which images to test.

build.json
```diff
@@ -5,6 +5,12 @@
     },
     "AIIDA_VERSION": {
       "default": "2.1.2"
+    },
+    "AIIDALAB_VERSION": {
+      "default": "22.08.0"
```
unkcpz, line 10: I made the new release `22.11.0`, can be bumped.

unkcpz (approved):
@danielhollas Thanks! Looks all good, just bump the `aiidalab` package version and I'll merge it.

danielhollas:
@unkcpz thanks. I've bumped the version. I think once the build finishes we should test it locally before merging, and make sure the packages are indeed getting installed to `~/.local/`

unkcpz:
Seems we need also release conda-fore of aiidalab, or we need to change the install method to pip.

danielhollas:
Ugh, right. I guess that's another strike against conda and plus for using pip to install aiidalab and aiida-core.

unkcpz (approved):
Tested with installing AWB, all good! thanks @danielhollas

PR #331, opened 2022-11-25
Using `notebook` CMD to run the jupyter stack

unkcpz:
The new `jupyter/minimal-notebook` image uses `lab` CMD to run the jupyter server, which was introduced by https://github.com/jupyter/docker-stacks/pull/1575. It provide `DOCKER_STACKS_JUPYTER_CMD` to set how to run it.
The lab (full-stack) is for AiiDAlab only and should start with notebook mode. I pin it as an environment variable to start the jupyter backend.
Since we have the new jupyterhub, the `NOTEBOOK_ARGS` can be used to set the notebook arguments.

### DIFF
--- a/stack/lab/Dockerfile
+++ b/stack/lab/Dockerfile
@@ -3,11 +3,11 @@ FROM base
 
 LABEL maintainer="AiiDAlab Team <aiidalab@materialscloud.org>"
 
-# The lab mode is default to run jupyter server.
-# Here the CMD pinned to `notebook` for running in legacy notebook backend
-# In the future, we have appmode compatible with nbclassic
-# we can use the server backend which is the alternative of legacy 
-# notebook backend.
+# By default, Jupyter Docker image launches the JupyterLab interface.
+# Here, we change it to the classic Jupyter Notebook which is used by AiiDAlab.
+# In the future, we might want to switch to other options such as `nbclassic` or `retro`, 
+# but the `nbclassic` is not supported because of appmode.
+# see: https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#switching-back-to-the-classic-notebook-or-using-a-different-startup-command
 ENV DOCKER_STACKS_JUPYTER_CMD=notebook
 
 USER root
--- a/stack/lab/Dockerfile
+++ b/stack/lab/Dockerfile
@@ -5,7 +5,7 @@ LABEL maintainer="AiiDAlab Team <aiidalab@materialscloud.org>"
 
 # By default, Jupyter Docker image launches the JupyterLab interface.
 # Here, we change it to the classic Jupyter Notebook which is used by AiiDAlab.
-# In the future, we might want to switch to other options such as `nbclassic` or `retro`, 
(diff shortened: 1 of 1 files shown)


danielhollas:
I tried to give it a test, but I am getting a 500 Jupyter server error, due to this jinja exception

```
      File "/opt/conda/lib/python3.9/site-packages/nbconvert/__init__.py", line 4, in <module>
        from .exporters import *
      File "/opt/conda/lib/python3.9/site-packages/nbconvert/exporters/__init__.py", line 3, in <module>
        from .html import HTMLExporter
      File "/opt/conda/lib/python3.9/site-packages/nbconvert/exporters/html.py", line 12, in <module>
        from jinja2 import contextfilter
    ImportError: cannot import name 'contextfilter' from 'jinja2' (/opt/conda/lib/python3.9/site-packages/jinja2/__init__.py)
```

This is probably another side-effect of updating the base image to Python 9.13? Perhaps we need to adjust the nbconvert / jinja dependencies, not sure where they are defined. 

https://github.com/d2l-ai/d2l-book/issues/46

danielhollas:
Ah, I see the error comes from AWB which reinstalls some dependencies and breaks the environment, which you're already trying to address in https://github.com/aiidalab/aiidalab-widgets-base/pull/395

If I don't install AWB, things seems to work.

stack/lab/Dockerfile
```diff
@@ -3,6 +3,13 @@ FROM base
 
 LABEL maintainer="AiiDAlab Team <aiidalab@materialscloud.org>"
 
+# The lab mode is default to run jupyter server.
```
danielhollas, line 6:
```suggestion
# By default, Jupyter Docker image launches the JupyterLab interface.
# Here, we change it to the classic Jupyter Notebook which is used by AiiDAlab.
# In the future, we might want to switch to other options such as 'nbclassic' or 'retro', see:
# https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#switching-back-to-the-classic-notebook-or-using-a-different-startup-command
```

stack/lab/Dockerfile
```diff
@@ -3,6 +3,13 @@ FROM base
...
+# Here the CMD pinned to `notebook` for running in legacy notebook backend
+# In the future, we have appmode compatible with nbclassic
```
danielhollas, line 8: Do I understand correctly that Appmode is currently not compatible with the `nbclassic` option?

danielhollas (approved):
I have verified that the issue exist on the current `aiidalab/full-stack:edge` and verified that this PR fixes it.

stack/lab/Dockerfile
unkcpz: Thanks! changed.

unkcpz:
> Ah, I see the error comes from AWB which reinstalls some dependencies and breaks the environment, which you're already trying to address in https://github.com/aiidalab/aiidalab-widgets-base/pull/395

Yes, exactly! @danielhollas thanks a lot for the review. I addressed the things you mentioned.

PR #333, opened 2022-11-26
Pre-install the gcc compiler to lab stack

unkcpz:
The main purpose to include gcc is to install `pymatgen` by pip, otherwise the build will fail.

### DIFF
--- a/stack/lab/Dockerfile
+++ b/stack/lab/Dockerfile
@@ -13,10 +13,13 @@ RUN apt-get update --yes && \
      && apt-get clean \
      && rm -rf /var/lib/apt/lists/*
 
-# Install aiidalab package
+ARG AIIDA_VERSION
+
+# Install aiidalab and aiida-core.atomic_tools
 ARG AIIDALAB_VERSION=22.08.0
 RUN mamba install --yes \
      aiidalab=${AIIDALAB_VERSION} \
+     aiida-core.atomic_tools=${AIIDA_VERSION} \
      && mamba clean --all -f -y && \
      fix-permissions "${CONDA_DIR}" && \
      fix-permissions "/home/${NB_USER}"


danielhollas:
@unkcpz how much does this change increase the size of the image?

unkcpz:
> @unkcpz how much does this change increase the size of the image?

Good question. The size increased quite a lot, the uncompressed `full-stack` image after pull increased from 2.22GB to 3.09GB. 
However, I think the packages in `aiida-core[atomic_tools]` is needed by aiidalab-widgets-base anyway. If the pymatgen can get installed easily by pip, I also don't want to make this workaround 😅, to be honest, I am very sick of the pymatgen.. We have one feature relying on it in QeApp to check the dimension of the structure. I think we get rid of it by removing the feature.

unkcpz:
Just google around and found two interesting tools for slim docker image:
1. https://github.com/docker-slim/docker-slim
2. https://github.com/wagoodman/dive

They can make the images multiple times thinner, deserve a try and then we can not too much care about the size.

danielhollas:
I would really like to avoid extra 800Mb just for pymatgen. I specifically
got rid of it in aiida-orca.

I wonder if there is a way to pin its version without installing it?

Dne so 26. 11. 2022 1:54 odp. uživatel Jusong Yu ***@***.***>
napsal:

unkcpz:
> I would really like to avoid extra 800Mb just for pymatgen.

Agree, don't want to yield to `pymatgen`. 

> I specifically got rid of it in aiida-orca.

Okay, nice! But the dependency is from aiida-quantumespresso plugin side, and it not very easy to get rid of it. 

The other option is to install `gcc` and then the pymatgen can be built without a problem. I test it the GCC increase `full-stack` from 2.22GB to 2.62GB.

> I wonder if there is a way to pin its version without installing it?

What is your purpose of this? The problem I faced is the plugin need it but the `pymatgen` can not installed with pip.

unkcpz:
Quick test with adding `pymatgen` only, it increase the size to 3.09GB directly, which means all size increase of `aiida-core[atomic_tools]` is from it 🤷🏻‍♂️

danielhollas:
> The problem I faced is the plugin need it but the pymatgen can not installed with pip.

Could you please open a separate issue and describe in detail what the problem is?
Or is it the one that you opened in aiida-core? https://github.com/aiidateam/aiida-core/issues/5767

When I ran into this issue, I was able to solve it by installing pymatgen via pip separately from the `aiida-core[atomic_tools]`. 
```
pip install -U pymatgen
```

Which suggests there is a way to install it via pip we just need to ensure it's the correct version.

As an alternative, could you try installing `aiida-core` and `aiidalab` via pip? This is what you suggested in #316, and perhaps that is the way to go because we always have this fundamental mismatch that aiida-core is installed by conda, by AiiDAlab Apps' dependencies are installed via pip.

danielhollas (changes requested):
I think this change needs more discussion and exploration of possible solutions. Perhaps we could discuss on Tuesday meeting?

unkcpz:
> I think this change needs more discussion and exploration of possible solutions. Perhaps we could discuss on Tuesday meeting?

Yes, I think we need a discussion, let's do it tomorrow.

unkcpz:
The failed build of `pymatgen` is manifested at https://github.com/aiidalab/aiidalab-docker-stack/actions/runs/3568133449/jobs/5999401435.

unkcpz:
The decision we made is to provide the `GCC` and `vim` in the `base` stack. I would still argue to install gcc through conda but vim can be install using `apt`. Since apt has less control of the version and gcc compiler should be very careful with the version. @danielhollas @yakutovicha if you don't have objections, I'll make a new PR for that.

unkcpz:
For `vim` it is not a problem, they have the `vim.tiny` installed in the new `minimal-notebook` image as the default editor.

unkcpz:
closed and re-addressed by #338

PR #335, opened 2022-11-28
Test AWB installation in CI

danielhollas:
Basic smoke tests to ensure we can install AWB and QeApp in the image. This is useful to identify conflicting dependencies downstream.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,7 +106,31 @@ jobs:
               shell: bash -l {0}  # required to activate the conda environment
               env: ${{ fromJSON(needs.build.outputs.images) }}
               run: |
-                  pytest -v
+                  pytest -v -m "not integration"
+
+    integration-test:
+        needs: test
+
+        runs-on: ubuntu-latest
+        timeout-minutes: 10
+
+        steps:
+            - uses: actions/checkout@v2
+            - name: Login to GitHub Container Registry
+              uses: docker/login-action@v2
+              with:
+                  registry: ghcr.io
+                  username: ${{ github.actor }}
+                  password: ${{ secrets.GITHUB_TOKEN }}
+            - name: Setup Python test environment
+              uses: mamba-org/provision-with-micromamba@v13
+              with:
+                  cache-env: true
+            - name: Run tests
+              shell: bash -l {0}  # required to activate the conda environment
+              env: ${{ fromJSON(needs.build.outputs.images) }}
+              run: |
+                  pytest -v -m integration
 
     release:
         if: >-
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,6 +108,7 @@ jobs:
               run: pytest -v -m "not integration"
 
     integration-test:
+        needs: build
         runs-on: ubuntu-latest
         timeout-minutes: 10
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,12 +105,9 @@ jobs:
             - name: Run tests
               shell: bash -l {0}  # required to activate the conda environment
               env: ${{ fromJSON(needs.build.outputs.images) }}
-              run: |
-                  pytest -v -m "not integration"
+              run: pytest -v -m "not integration"
 
     integration-test:
-        needs: test
-
         runs-on: ubuntu-latest
         timeout-minutes: 10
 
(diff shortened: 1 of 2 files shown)


danielhollas:
@unkcpz this is ready for review. The new test is failing for QeApp, see
https://github.com/aiidalab/aiidalab-docker-stack/actions/runs/3568133449/jobs/5996718920

```
E           INFO: pip is looking at multiple versions of <Python from Requires-Python> to determine which version is compatible with other requirements. This could take a while.
E           INFO: pip is looking at multiple versions of aiidalab-qe to determine which version is compatible with other requirements. This could take a while.
E           ERROR: Cannot install aiidalab-qe and aiidalab-qe==22.11.1 because these package versions have conflicting dependencies.
E         ? +++++
E           
E           The conflict is caused by:
E               aiidalab-qe 22.11.1 depends on aiida-quantumespresso~=4.1
E               aiidalab-qe-workchain 22.11.1 depends on aiida-quantumespresso~=3.2
E           
E           To fix this you could try to:
E           1. loosen the range of package versions you've specified
E           2. remove package versions to allow pip attempt to solve the dependency conflict
E           
E           ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
E           Error: Failed to install 'aiidalab-qe' (version=git+https://github.com/aiidalab/aiidalab-qe.git) at '/home/jovyan/apps/aiidalab-qe', due to error: Failed to install dependencies.

tests/test_aiidalab.py:130: AssertionError
```
So it seems dependencies for QeApp needs to be fixed as well, similarly to what you did for AWB.

unkcpz (changes requested):
@danielhollas thanks! This is an interesting and useful test. 
These two test can be combined by fixture or as pytest parameters, but I prefer they are separated as it is now, I'll let you decide. 

I have another concern that when we make changes to this repo, it is high potentially failing the test. I'd purpose separating this test out and run as a separated workflow, so we can set "required" branch protect to other tests but not strictly to this test.

unkcpz:
> So it seems dependencies for QeApp needs to be fixed as well, similarly to what you did for AWB.

Sure! I finish the migration but not make the release yet.

.github/workflows/ci.yml
```diff
@@ -105,8 +105,28 @@ jobs:
             - name: Run tests
               shell: bash -l {0}  # required to activate the conda environment
               env: ${{ fromJSON(needs.build.outputs.images) }}
-              run: |
-                  pytest -v
+              run: pytest -v -m "not integration"
+
+    integration-test:
+        runs-on: ubuntu-latest
```
danielhollas, line 111:
```suggestion
        needs: build
        runs-on: ubuntu-latest
```

danielhollas:
@unkcpz I've moved the tests to a separate workflow as you suggested. 

I have removed the aiidalab-qe test so that the test pass, but I expect that you will enable it in your GCC PR.

.github/workflows/ci.yml
```diff
@@ -105,8 +105,29 @@ jobs:
             - name: Run tests
               shell: bash -l {0}  # required to activate the conda environment
               env: ${{ fromJSON(needs.build.outputs.images) }}
-              run: |
-                  pytest -v
+              run: pytest -v -m "not integration"
```
unkcpz, line 108: How is this `-m` option work?

.github/workflows/ci.yml
```diff
@@ -105,8 +105,29 @@ jobs:
...
+                  username: ${{ github.actor }}
+                  password: ${{ secrets.GITHUB_TOKEN }}
+            - name: Setup Python test environment
+              uses: mamba-org/provision-with-micromamba@v13
+              with:
+                  cache-env: true
+            - name: Run integration tests
+              shell: bash -l {0}  # required to activate the conda environment
+              env: ${{ fromJSON(needs.build.outputs.images) }}
+              run: pytest -v -m "integration"
```
unkcpz, line 130: The name "integration" is a bit confuse because "I" for CI test is the meaning of integration. But I have no better suggestion TBH.

unkcpz (approved):
@danielhollas thanks! LGTM.

.github/workflows/ci.yml
danielhollas: https://docs.pytest.org/en/7.1.x/example/markers.html#marking-test-functions-and-selecting-them-for-a-run

(I find the not operator quite strange, but it works)

unkcpz: never mind. I find it, a useful option.

.github/workflows/ci.yml
danielhollas: I think the new tests fit the definition of 'integration tests' which is a standard term I think.

https://circleci.com/blog/unit-testing-vs-integration-testing/

PR #338, opened 2022-11-29
Add gcc compiler in base stack

unkcpz:
The gcc is installed in`build-essential` by apt. The size of `full-stack` increased from 2.22GB to 2.43GB which is half the size it installed by conda. 
I also test with installing the pymatgen and all went will.

### DIFF
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -7,7 +7,7 @@ USER root
 
 RUN apt-get update --yes && \
     apt-get install --yes --no-install-recommends \
-    # for apps which need to install pymatgen: 
+    # for apps which need to install pymatgen:
     # https://pymatgen.org/installation.html#installation-tips-for-optional-libraries
     build-essential && \
     apt-get clean && rm -rf /var/lib/apt/lists/*
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -4,6 +4,14 @@ FROM ${BASE}
 LABEL maintainer="AiiDAlab Team <aiidalab@materialscloud.org>"
 
 USER root
+
+RUN apt-get update --yes && \
+    apt-get install --yes --no-install-recommends \
+    # for apps which need to install pymatgen: 
+    # https://pymatgen.org/installation.html#installation-tips-for-optional-libraries
+    build-essential && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
+
 WORKDIR /opt/
 
 ARG AIIDA_VERSION
--- a/tests/test_aiidalab.py
+++ b/tests/test_aiidalab.py
@@ -117,7 +117,7 @@ def test_install_apps_from_stable(aiidalab_exec, package_name, nb_user, variant)
 
 
 @pytest.mark.integration
-@pytest.mark.parametrize("package_name", ["aiidalab-widgets-base"])
+@pytest.mark.parametrize("package_name", ["aiidalab-widgets-base", "aiidalab-qe"])
 def test_install_apps_from_master(aiidalab_exec, package_name, nb_user, variant):
     if "lab" not in variant:
         pytest.skip()


stack/base/Dockerfile
```diff
@@ -4,6 +4,14 @@ FROM ${BASE}
 LABEL maintainer="AiiDAlab Team <aiidalab@materialscloud.org>"
 
 USER root
+
+RUN apt-get update --yes && \
+    apt-get install --yes --no-install-recommends \
+    # for apps which need to install pymatgen:
+    # https://pymatgen.org/installation.html#installation-tips-for-optional-libraries
+    build-essential && \
```
danielhollas, line 12: I wonder if the whole `build-essential` package is needed? Did you try installing gcc directly?

unkcpz: I did not test but the `build-essential` is what has been moved from original minimal-notebook image. Same suggestion in https://github.com/jupyter/docker-stacks/issues/1385#issuecomment-868379259

unkcpz: Also seems when google "ubuntu install gcc", it is recommended to install the `build-essential`. I guess some libraries are all needed and should be installed at the same time.

yakutovicha (commented):
for your information, this is the full list of packages that was installed on aiida-prerequisites: https://github.com/aiidateam/aiida-prerequisites/blob/main/Dockerfile#L47-L66

stack/base/Dockerfile
yakutovicha: I found something here: https://itslinuxfoss.com/install-gcc-ubuntu-22-04/. They do `sudo apt install gcc-12`

But indeed, by default, they all recommend installing `build-essential`. It contains the "essential tools for development".

unkcpz:
Thanks! I think some of them are in the `minimal-notebook`, we can add them back on requirements depending on the asks by users.

stack/base/Dockerfile
danielhollas: @unkcpz could you maybe just try installing just gcc, to see how the size would be different?

> we do need to run make

Do we? (I do not know, just curious where it is needed)

unkcpz:
The vim can start by `$ vi`, why you said it is hard to use? @danielhollas

stack/base/Dockerfile
yakutovicha: Anyways, I guess the easiest thing to do here would be to install gcc and try to install numpy/pymatgen. If that fails, probably build-essential is the best option. @unkcpz can you try, please?

danielhollas: @unkcpz what was the size difference when you only installed gcc?

Those are just missing header files, so I'd say installing `libc6-dev` is safe and I would go for it if there is a meaningful size difference from fulll `build-essential` package. If the difference is small, let's just install the whole thing.

unkcpz: current `full-stack` -> 2.22GB
whole `build-essential` -> 2.43GB
only `gcc` + `libc6-dev` -> 2.38GB

I'd vote to have `build-essential`, the size already increase, so make sense for extra ~50MB for all toolchains.

danielhollas:
> The vim can start by `$ vi`, why you said it is hard to use? @danielhollas

When vim runs in the so called compatible mode it behaves very strangely and many ViM improvements over Vi are disabled. For example, try searching with `\n` in the compatible mode..

It looks like just having an empty `/home/jovyan/.vimrc` will automatically set the non-compatible mode, so that is what I propose, per: https://superuser.com/a/543327

unkcpz:
> It looks like just having an empty /home/jovyan/.vimrc will automatically set the non-compatible mode, so that is what I propose, per: https://superuser.com/a/543327

Sounds good. Let's do it in a separate PR. I'll do it.

stack/base/Dockerfile
yakutovicha: I agree with @unkcpz to go for build-essential 👍

danielhollas (approved):
Okay, let's go with `build-essential`. Thanks @unkcpz for the additional tests and analysis. :pray: 

I am still not stoked about the size increase but cannot think of a better solution. Maybe if somebody attends the aiida-core meetings you could bring up the issue with them in case anybody has better idea.

unkcpz:
>  Maybe if somebody attends the aiida-core meetings you could bring up the issue with them in case anybody has better idea.

Sure, I attend the meeting. but the issue is about pymatgen, I don't think there is a workaround on our side 😁, just another complaint about `pymatgen`.

danielhollas:
@unkcpz FYI in this PR that adds Python-3.11 support to aiida-core, @sphuber is updating the pymatgen version in `aiida-core[atomic-tools]`. It looks like it might resolve some issues with compatibility with numpy, although I guess we still need `gcc` to build it?

sphuber:
I am indeed working on this now and this should hopefully resolve the problems we have been having since as off that version they finally provided an explicitly compatible version of `numpy` in the build requirements. This should solve: https://github.com/aiidateam/aiida-core/issues/5767

I am now finalizing the PR and it should be almost done. It was initially blocked since I added Python 3.11 support to the latest versions of `kiwipy` and `plumpy`, but those include a lot more other changes that require a lot more work in `aiida-core` and so it was blocked. But today I have backported the support to the support branches and made new releases, so we should now be able to merge this in `main` and release it with `aiida-core==2.2.0`.

chrisjsewell:
Guys, why are you installing anything via pip? If you have mamba installed then you should be installing aiida-quantume-espresso and thus pymatgen via mamba: https://github.com/conda-forge/aiida-quantumespresso-feedstock

This completely negates any issues with numpy 🤷

chrisjsewell:
Also, in the Docker you have:

```
RUN mamba install --yes \
     aiida-core=${AIIDA_VERSION} \
     --file requirements.txt
```

then in the requirements.txt you have:

```
aiida-core>=2.0.0,<3
pip==22.0.4
```

that doesn't seem to make any sense?

danielhollas:
Hi @chrisjsewell and @sphuber,

thanks for chiming in!

> Guys, why are you installing anything via pip?

The way AiiDAlab apps are installed is as follows:

1. Clone the git repo with the App (e.g. https://github.com/aiidalab/aiidalab-qe)
2. Install its dependencies by running `pip install .` inside the repo.

I don't think this workflow is doable with conda / mamba, but LMK if I am mistaken. There are certainly many drawbacks of doing it this way.

@unkcpz  It does make me think though, we also have a `post_install` script that is run automatically after the pip installation. We use this for example to install quantum-espresso itself inside a new conda environment. So we could in principle remove the dependency on the `aiida-quantum-espresso` plugin from `setup.cfg` and install it into the default conda environment in `post_install`. But I don't know if that is a sustainable solution. We would need to assure there's not overlap in dependencies, and this solution would need to be adopted by other AiiDAlab apps.

> then in the requirements.txt you have:

I don't see a problem here? Note that the `requirements.txt` file is also used as the pip constraint file to prevent Apps breaking the basic environment.

> https://github.com/aiidalab/aiidalab-docker-stack/blob/6036e06a3dc91ed07272ebc184197493e0984164/stack/base/Dockerfile#L33

chrisjsewell:
> is also used as the pip constraint file

Ok fair, just seems a bit off to have an `AIIDA_VERSION` which can then actually be overridden (within the constraints) by an app

> Install its dependencies by running `pip install .` inside the repo.

yeh mixing conda and pip installs is definitely not recommended

chrisjsewell:
See e.g. https://www.anaconda.com/blog/using-pip-in-a-conda-environment

If not already, at the least you should be using: `pip install --upgrade-strategy only-if-needed`

unkcpz:
@chrisjsewell thanks a lot for the suggestion. An extra comment based on what was mentioned by @danielhollas, in AiiDAlab, the app is not needed to be published to PyPI or conda which reduces the burden for app developers. 

> yeh mixing conda and pip installs is definitely not recommended

Fair, that's the reason I propose to use conda to only install packages that need to be compiled, the python packages should installed by pip. The mamba/conda in the context is mostly used to create the environment. We see a lot of unsolved issues that the conda doesn't have the latest package in conda-forge. 

> If not already, at the least you should be using: pip install --upgrade-strategy only-if-needed

This is very helpful, and I think we need this to put the constraint on the app install.

danielhollas:
> If not already, at the least you should be using: `pip install --upgrade-strategy only-if-needed`

@chrisjsewell I am looking at the [pip documentation](https://pip.pypa.io/en/stable/development/architecture/upgrade-options/#controlling-what-gets-installed), which says about `--upgrade-strategy`

> This option affects which packages are allowed to be installed. It is only relevant if --upgrade is specified 

So in our case where we do not run with `--upgrade` it would have no effect?

> Fair, that's the reason I propose to use conda to only install packages that need to be compiled, the python packages should installed by pip. The mamba/conda in the context is mostly used to create the environment.

Yeah, I've opened a new PR #336 where I use pip to install `aiida-core` and `aiidalab`. The basic tests are passing so perhaps that is the way to go in the future.

chrisjsewell:
> where I use pip to install aiida-core and aiidalab. 

Personally I would go completely the opposite way, but up to you guys

danielhollas:
> in AiiDAlab, the app is not needed to be published to PyPI or conda which reduces the burden for app developers.

I think this point is quite important, if want to people to be able to develop AiiDAlab apps easily. I've never published anything on conda-forge, and I am sure it is not that hard in the end, but at least looking from the distance the whole process (and even just the conda-forge nomenclature) seems non-trivial. In any case, this would be a non-trivial and breaking change so probably cannot happen any time soon, unless we want to do it now before the 2.0 compatibility release.

From the checklist on Anaconda article:

> :heavy_check_mark: Use pip only after conda
   - :heavy_check_mark:  install as many requirements as possible with conda, then use pip
   - :heavy_check_mark: pip should be run with --upgrade-strategy only-if-needed (the default)
   - :negative_squared_cross_mark:  Do not use pip with the --user argument, avoid all “users” installs

The last point we do not satisfy, and we've just changed it because we needed to persist the app dependencies in the mounted volume. :expressionless:

PR #342, opened 2022-12-01
Install povray

danielhollas:
Povray is needed by AWB to render molecular images. Currently the render button in the Download tab of StructureViewer is broken because of this missing dependency. povray must be installed by apt, so needs to be installed in the image.

CC @unkcpz

### DIFF
--- a/stack/lab/Dockerfile
+++ b/stack/lab/Dockerfile
@@ -16,7 +16,8 @@ WORKDIR /opt/
 # Install additional system packages
 RUN apt-get update --yes && \
      apt-get install --yes --no-install-recommends \
-     curl povray \
+     curl \
+     povray \
      && apt-get clean \
      && rm -rf /var/lib/apt/lists/*
 
--- a/stack/lab/Dockerfile
+++ b/stack/lab/Dockerfile
@@ -16,7 +16,7 @@ WORKDIR /opt/
 # Install additional system packages
 RUN apt-get update --yes && \
      apt-get install --yes --no-install-recommends \
-     curl \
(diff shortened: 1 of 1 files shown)


stack/lab/Dockerfile
```diff
@@ -16,7 +16,7 @@ WORKDIR /opt/
 # Install additional system packages
 RUN apt-get update --yes && \
      apt-get install --yes --no-install-recommends \
-     curl \
+     curl povray \
```
unkcpz, line 19:
```suggestion
     curl \
     povray \
```

unkcpz: And why the `curl` is needed? I think `curl` is useful and cat go to `base`.

unkcpz (changes requested):
@danielhollas thanks! I am surprised that this is not a python package and has to be installed in the system. Just a minor request.

stack/lab/Dockerfile
danielhollas: I am not sure why it is here. I'd rather keep the base basic and slim so keeping it here.

unkcpz:
Ah, it was already there. I thought you add it. Nevermind, all good.

unkcpz (approved):

danielhollas:
@unkcpz you might want to do a new docker-stack release to fix the render button in AWB before you do public testing.

unkcpz:
@danielhollas thanks for the head up! I find some other small issues during tests these days, I'll see if I can quick fix them tomorrow, then I'll make the release.

Issue #320, opened 2022-10-18
Pymatgen installed via aiida-core[atomic_tools] does not work in full-stack image


danielhollas:
Try installing `aiida-core[atomic_tools]` in the new `aiidalab/full-stack`
```console
pip install aiida-core[atomic_tools]
```
(that's what happens when I install `aiida-orca` plugin as a dependency in my AiiDAlab app).

Then try the following import (taken from AWB)
```python
>>> from pymatgen.core import Structure, Molecule
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/conda/lib/python3.9/site-packages/pymatgen/core/__init__.py", line 20, in <module>
    from .lattice import Lattice  # noqa
  File "/opt/conda/lib/python3.9/site-packages/pymatgen/core/lattice.py", line 22, in <module>
    from pymatgen.util.coord import pbc_shortest_vectors
  File "/opt/conda/lib/python3.9/site-packages/pymatgen/util/coord.py", line 17, in <module>
    from . import coord_cython as cuc
  File "pymatgen/util/coord_cython.pyx", line 1, in init pymatgen.util.coord_cython
ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject

```

unkcpz:
Will update numpy to `1.21` manually solve the problem?

danielhollas:
One problem might be that numpy is installed via conda from `conda-forge` (I guess as part of AiiDA dependencies) while pymatgen gets installed via pip in my case. Reinstalling numpy via pip seems to solve the issue.

```
pip install -U numpy
```

unkcpz:
Thanks for checking, @danielhollas. 
aiida-core pinned the [`numpy` version to `1.19`](https://github.com/aiidateam/aiida-core/blob/025852270ee30f6249420645eddab5ad0e0ca136/pyproject.toml#L41). I think that's is the cause of the issue when installing the `pymatgen` from `aiida-core[atomic_tools]`. Could you open an issue to the `aiida-core`? Bumping the `numpy` version solve the issue, but may bring other problem to other dependencies.

danielhollas:
@unkcpz I just noticed this line in the setup.cfg in the aiidalab-qe repo

```
    numpy~=1.23 # pined for pymatgen
```
I guess that above error might be the reason for this?

unkcpz:
[...]

PR #345, opened 2022-12-13
Bump aiida-core version to 2.2.1

danielhollas:
New version of aiida-core has been released today, with a couple of important bugfixes (one of them I might need for my App, not sure). Among others it also has updated pymatgen dependency that should prevent some of the issues we've had, as described in #320.


yakutovicha (approved):
LGTM!

Issue #343, opened 2022-12-06
Well management of conda/pip hybrid python environment


unkcpz:
Move the post-PR discussion of conda/pip hybrid environment https://github.com/aiidalab/aiidalab-docker-stack/pull/338#issuecomment-1331199302 as an issue here for further discussion.

Summary:
- Not a very good idea to [mix use pip in a conda environment](https://www.anaconda.com/blog/using-pip-in-a-conda-environment)
- Have to do it at the moment in AiiDAlab since we want to make it is possible that app can be install without publishing the PyPI or conda-forge. This reduces the burden for app developers.
- The principle is conda only for packages that need to be compiled (how we know this in advance, for instance, the `pymatgen`) and pip for or regular python packages.

danielhollas:
Some other things to consider

1. Currently, one can install conda dependencies via the `post_install` script that runs automatically when the app is installed. However, although in the new stack it is now possible to install into the `base` environment in `/opt/conda`, this doesn't really work since these dependencies would not persist when the container exits.
2. One can create a new conda environment, but than the dependencies are not available to the jupyter server, which runs under the `base` environment. In QeApp, a new conda environment is used for installing Quantum Espresso, and this environment is activated in the `prepend_text` when the AiiDA daemon runs the calculations. But how do we install conda dependencies that should be available in the AiiDAlab app itself? This is a problem I am also facing in my App.
3. As we saw with AWB dependencies, pip will happily downgrade packages when installing into `~/.local` directory, which may break the environment. We probably need to use the pip constraint file (`requirements.txt`) more, at least we should constrain the Jupyter dependencies.

danielhollas:
Given the above, a counter proposal would be to use conda for most of the stuff. 

I think this could be done if we cloned the `base` environment into `~/.conda` at first startup (we could call it `local_base`), and activate this new environment before the Jupyter server is started. Then the apps could install into this new base environment and these dependencies would be available to Jupyter server.

Instead of Apps having dependencies in `setup.cfg`, we would have them in a separate file, e.g. `conda-environment.txt` and they would be automatically installed by conda, which is much better at ensuring environment consistency than pip. In this world, we would only use pip to build and install the AiiDAlab application itself.

danielhollas:
Possibly relevant, how to run ipython kernels from different conda environments.

https://towardsdatascience.com/get-your-conda-environment-to-show-in-jupyter-notebooks-the-easy-way-17010b76e874

@yakutovicha

[...]

PR #346, opened 2022-12-13
Create aiida-base environment and clone it to home folder at startup

danielhollas:
Testing an approach for resolving #343

### DIFF
--- a/build.json
+++ b/build.json
@@ -7,7 +7,7 @@
       "default": "2.1.2"
     },
     "AIIDALAB_VERSION": {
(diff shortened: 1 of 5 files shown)


danielhollas:
Tests currently fail due to unrelated issue with `packaging` in `aiidalab` fixed in https://github.com/aiidalab/aiidalab/pull/341 but not yet released.

danielhollas:
This was an interesting investigation but I don't think it's workable in the end so closing for now

Issue #347, opened 2022-12-14
bashrc disappear in the k8s deployment


unkcpz:
In the THEOS deployment, it is weird that the `.bashrc` is not there after the server start. @danielhollas do you have the same issue in your development?
Luckily most of the things work fine, just the verdi completion is not working at the moment since the `base` conda environment is not activated.

danielhollas:
I haven't seen this issue, but I am only using aiidalab-launch

unkcpz:
Yes, the `aiidalab-launch` has no issue.

unkcpz:
I think I find the catch of this issue. Running the container locally by `docker run` or using `aiidalab-launch` will start the container by `start-notebook.sh`, while starting container each user by spawned from jupyterhub on k8s and ansible deployment the container is started by `start-singleuser.sh`. The difference can be find:

- `start-singleuser.sh`: https://github.com/jupyter/docker-stacks/blob/10e52ee84369aaa37630aea13c137b766ace307b/base-notebook/start-singleuser.sh
- `start-notebook.sh`: https://github.com/jupyter/docker-stacks/blob/10e52ee84369aaa37630aea13c137b766ace307b/base-notebook/start-notebook.sh

This explains the issues I and @yakutovicha encountered that the environment variable `DOCKER_STACKS_JUPYTER_CMD ` set in the `lab` stack didn't take effect. 
I need more investigation where the bash profile is set or wiped.

unkcpz:
Here is a relevant discussion how to properly using start-singleuser.sh. https://github.com/jupyter/docker-stacks/issues/1528

unkcpz:
I did some investigation and think the problem caused from that the k8s deployment (also the ansible deployment) where the home is mounted after the spawner will wipe the configurations setted such as the `.bashrc` and `.vimrc`. It mentioned in the issue https://github.com/jupyter/docker-stacks/issues/815#issuecomment-478363912. 

This means in our docker stack we need to avoid modifying the `~/.bashrc` directly but `/etc/skel/bashrc` and copy it to `~/.bashrc` afterward in `before-notebook.d`.

PR #348, opened 2022-12-30
New conda environments in ~/.conda/ by default

### LINKED ISSUE #352: New conda environments are created in `/opt/conda` by default
In the current image, when a user creates a new conda environment with `conda create --name new-env`, it will be created in `/opt/conda/envs`, but that's no good since it will not be persisted between container invocation. 

In the QeApp, this is solved by specifying the full path to the environment like this:

```sh
conda create -p ~/.conda/envs/quantum-espresso
```

https://github.com/aiidalab/aiidalab-qe/blob/d6b6d046d42b186be60bbd34ca8d001a3fc4060a/aiidalab_qe/setup_codes.py#L44
But the disadvantage is that to activate this environment, one needs to provide the full path as well, which is clumsy.

I think we should modify the conda config so that new environments are created in `~/.conda/envs/` by default.
This change would actually make the behavior consistent with the old docker stack image so it should make migration easier for AiiDAlab app developers.

danielhollas:
In the current image, when a user creates a new conda environment with `conda create --name new-env`, it will be created in `/opt/conda/envs`, but that's no good since it will not be persisted between container invocation. 

In the QeApp, this is solved by specifying the full path to the environment like this:

```sh
conda create -p ~/.conda/envs/quantum-espresso
```

https://github.com/aiidalab/aiidalab-qe/blob/d6b6d046d42b186be60bbd34ca8d001a3fc4060a/aiidalab_qe/setup_codes.py#L44
But the disadvantage is that to activate this environment, one needs to provide the full path as well, which is clumsy.

In this PR I modify the conda config so that the new environments are created in `~/.conda/envs/` by default.
This change actually makes the behavior consistent with the old docker stack image, so it should make migration easier for AiiDAlab app developers.

Closes #352

### DIFF
--- a/stack/base-with-services/Dockerfile
+++ b/stack/base-with-services/Dockerfile
@@ -8,7 +8,7 @@ WORKDIR /opt/
 
 ARG AIIDA_VERSION
 
-RUN mamba create -n aiida-core-services --yes \
+RUN mamba create -p /opt/conda/envs/aiida-core-services --yes \
      aiida-core.services=${AIIDA_VERSION} \
      rabbitmq-server=3.8.14 \
      && mamba clean --all -f -y && \
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -63,6 +63,13 @@ fi\n' >> "/home/${NB_USER}/.bashrc"
 # Add ~/.local/bin to PATH where the dependencies get installed via pip
 ENV PATH=${PATH}:/home/${NB_USER}/.local/bin
 
+# Add conda envs_dirs in home directory,
+# which will persist between container invocation
+# NOTE: The order here is important!
+# We want conda to create environments in ~/.conda/ by default
+RUN conda config --system --add envs_dirs /opt/conda
+RUN conda config --system --add envs_dirs "~/.conda/envs"
+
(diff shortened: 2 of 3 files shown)


yakutovicha (approved):
LGTM!

unkcpz:
Just a reminder, we need to be careful to add more configurations to home folder. As mentioned in https://github.com/aiidalab/aiidalab-docker-stack/issues/347#issuecomment-1370901231, when starting the image by singleuser.sh by spawner in production deployment, the mounted home volume will be wiped.

unkcpz:
> Unrelated to that, keeping the conda environment in the home folder seems the right choice.

My concern is even in this PR the `~/.conda` is added to the home folder, I think is the production development by Jupyterhub, the files will not exist after the container started. 
Instead, everything added to the HOME should add by a script in `before-notebook.d`.

unkcpz:
Sorry, I was wrong about this, I didn't look carefully at the change. This PR didn't make any modifications to the home folder.

danielhollas:
@unkcpz indeed.Here I only modified the global conda configuration, which is itself stored I believe in `/opt/conda/.condarc`

PR #350, opened 2023-01-01
Fix .vimrc placement

danielhollas:
We cannot create the `.vimrc` file in the Dockerfile, because the home folder gets mounted. We need to create it on startup.


danielhollas:
Ah, the file is there, it's just that the empty vimrc doesn't seem to have the same effect as having the 'set nocp' there explicitly. Since nobody else is complaining I'll close this for now.

PR #353, opened 2023-01-04
Move rc configures in home folder to before-notebook

### LINKED ISSUE #347: bashrc disappear in the k8s deployment

unkcpz:
fixes #347 

If the container is started by spawner and the home is remounted.
The .bashrc in HOME won't be set properly. I generate the `.bashrc`, `.profile` and `.bash_logout` from `/etc/skel/` folder and move the bashrc modification for singlesshagent into the before-notebook script.

This is tested in the THEOS k8s deployment. However, the issue manifest only when start by singleuser, the CI test will not touch the changed code.


stack/base/before-notebook.d/10_prepare-home-rc.sh
```diff
@@ -0,0 +1,23 @@
...
+if [[ ! -f  /home/${NB_USER}/.bash_logout ]]; then
+    cat /etc/skel/.bash_logout > /home/${NB_USER}/.bash_logout
+fi
+
+# The vim.tiny shipped with jupyter base stack start vim
+# in compatible mode, by creating `.vimrm` file in user home
+# make it nocompatible mode so the modern vim features is available:
+# https://superuser.com/questions/543317/what-is-compatible-mode-in-vim/543327#543327
+if [[ ! -f  /home/${NB_USER}/.vimrc ]]; then
+    touch .vimrc
```
danielhollas, line 22: Since we're touching this, can we change this to 

```suggestion
    echo "set nocp" > .vimrc
```

I've noticed it makes a difference for me (without this I sometimes still get weird behaviour).

```

File: stack/base/before-notebook.d/10_prepare-home-rc.sh [hunk 2]
```diff
@@ -0,0 +1,23 @@
...
+if [[ ! -f  /home/${NB_USER}/.profile ]]; then
+    cat /etc/skel/.profile > /home/${NB_USER}/.profile
+fi
+
+if [[ ! -f  /home/${NB_USER}/.bash_logout ]]; then
+    cat /etc/skel/.bash_logout > /home/${NB_USER}/.bash_logout
+fi
+
+# The vim.tiny shipped with jupyter base stack start vim
+# in compatible mode, by creating `.vimrm` file in user home
```
danielhollas, line 18:
```suggestion
# in compatible mode, by creating `.vimrc` file in user home
```

stack/base/before-notebook.d/10_prepare-home-rc.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# If the container is start by spawner and the home is remounted.
+# The .bashrc in HOME won't be set properly.
+if [[ ! -f  /home/${NB_USER}/.bashrc ]]; then
+    cat /etc/skel/.bashrc > /home/${NB_USER}/.bashrc
```
danielhollas, line 6: Why not 
```suggestion
    cp /etc/skel/.bashrc > /home/${NB_USER}/.bashrc
```

(and same below). Seems weird to use `cat` for copying a file.

stack/base/before-notebook.d/10_prepare-home-rc.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# If the container is start by spawner and the home is remounted.
```
danielhollas, line 3: Can you explain a bit more how is this home remounting different from the mounting that we do normally e.g. when using `aiidalab-launch`? 
Does this mean that there's a pre-existing `.bashrc` in the mounted volume that overrides the one from the image? Or something else?

danielhollas (changes requested):
Couple small things and a clarifying question. I am not familiar with the kubernetes deployment, so I'd like to understand more why this issue occurs only there.

stack/base/before-notebook.d/10_prepare-home-rc.sh
unkcpz: I have no exact clue why it is overrided. I suspect the order of creating persistence volume after the container is created matters. I'll try to do more investigation see if there any document mentioned the order of mounting order. I need to compare with aiidalab-launch why there is no issue there since it creates volume and mount but not these issues.

yakutovicha: Could it be due to the home mount configuration for the Kubernetes cluster?

stack/base/before-notebook.d/10_prepare-home-rc.sh
giovannipizzi: If you use `cp`, you need to remove the `>` sign

unkcpz:
Hi @yakutovicha, could you give it another look?

yakutovicha:
Sorry, I didn't have time to look at it yesterday. Two things:

1. I assume both Postgres and AiiDA databases are stored inside the home directory. We should handle them too.
2. Let's create an issue to find a way to configure spawners to mount the home directory together with copying the data.

unkcpz:
No worries, I should apologize for merging without your approve. @yakutovicha

> I assume both Postgres and AiiDA databases are stored inside the home directory. We should handle them too.

I don't think so, we don't have problem with that since the AiiDA database is created by the the script https://github.com/aiidalab/aiidalab-docker-stack/blob/main/stack/base/before-notebook.d/40_prepare-aiida.sh.

We only need to concern about files and folders that are created in the image by the Dockerfile.

> Let's create an issue to find a way to configure spawners to mount the home directory together with copying the data.

Not sure for different spawners there is a general way to do it. If we need to copy some files to a persistent volume in the future, I don't think the code can be added anywhere in this repository. It is better to go to the documentation on how to copy them.

danielhollas:
@unkcpz by the way the merge commit of this PR failed to build on main, with error

```
ERROR: failed to solve: failed to push ghcr.io/aiidalab/base-with-services:edge: failed to copy: io: read/write on closed pipe
```
I've noticed transient errors like this before, but I've tried to restart the CI job several times but it keeps failing so not sure what's happening. :shrug:

unkcpz:
I increase the timeout and the action re-triggered. Seems build all fine https://github.com/aiidalab/aiidalab-docker-stack/actions/runs/3930000098/jobs/6719466454

stack/base/before-notebook.d/20_setup-ssh.sh
```diff
@@ -11,5 +11,11 @@ if [[ ! -f  /home/${NB_USER}/.ssh/id_rsa ]]; then
   ssh-keygen -f /home/${NB_USER}/.ssh/id_rsa -t rsa -b 4096 -m PEM -N ''
 fi
 
+# Set by source load-singlesshagent.sh script
+echo $'\n# Load singlesshagent on shell startup.
```
danielhollas, line 15: @unkcpz I've noticen an issue with this code, the '.bashrc' gets appended everytime the container is started, so this code gets needlessly duplicated. I would proposed to move this in the `10_prepare-home-rc.sh`, where you can append it just one when the `.bashrc` is first copied over.

unkcpz: I think you are correct. But I initially didn't move this to `10_prepare-home-rc.sh` in order to keep all ssh related setting in one file and to make sure the command runs after running the ssh-keygen (which is not necessary but for readability).
Maybe instead of downloading the `load-singlesshagent.sh` on the fly, it is easy to manually maintain it in this repo and add a file existence check before running it? What you think?

danielhollas: @unkcpz I've created #357 to keep track of this issue. I assigned you and put to next sprint since I think it would be good to fix before the next docker stack release. Feel free to disagree though...

PR #356, opened 2023-02-01
Pin major PostgreSQL version in the image

### LINKED ISSUE #355: Pin PostgreSQL version
I've just hit an issue in one of my full-stack based container, which cannot start since it was created with PSQL=14, but somewhere along the way we updated to PSQL==15.

I've looked around and we don't seem to be in control of the PSQL version in our docker stacks. PSQL is installed as part of the `aiida-core.services` conda package. However, looking the the conda recipe on conda-forge, the psql version is not pinned, only a lower bound is given. @sphuber do I understand the situation correctly?

https://github.com/conda-forge/aiida-core-feedstock/blob/9bf266a74d8af2b0d866bb4d99f853fa3785c5c1/recipe/meta.yaml#L103

If that is the case, a major PSQL update can happen essentially any time we rebuild and publish the docker stack images here, which seems suboptimal, to put it mildly. Even when the automatic postgres migration is finished in #25, I'd say it is still highly desirable to be in control of the Postgres version, so that we do not inflict migrations on users randomly. 

@unkcpz @yakutovicha if you agree, I'll cook up a PR with the change.

danielhollas:
Pinning major PostgreSQL version in the Docker stack, as discussed in #355. I am not pinning minor version, since minor version updates shouldn't require DB migration.

I am also adding the `postgresql-{version}` docker tag to stable releases, in the same way we tag the images with Python and AiiDA versions. That should make it easy for users to avoid migrations if they need to.

Closes #355

### DIFF
--- a/build.json
+++ b/build.json
@@ -3,6 +3,9 @@
     "PYTHON_VERSION": {
       "default": "3.9.13"
     },
+    "PGSQL_VERSION": {
+      "default": "15"
+    },
     "AIIDA_VERSION": {
       "default": "2.2.1"
     },
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,6 +167,7 @@ jobs:
                       type=raw,value={{tag}},enable=${{ github.ref_type == 'tag' && ! startsWith(github.ref_name, 'v') }}
                       type=raw,value=aiida-${{ env.AIIDA_VERSION }},enable=${{ github.ref_type == 'tag'  && startsWith(github.ref_name, 'v') }}
                       type=raw,value=python-${{ env.PYTHON_VERSION }},enable=${{ github.ref_type == 'tag' && startsWith(github.ref_name, 'v') }}
+                      type=raw,value=postgresql-${{ env.PGSQL_VERSION }},enable=${{ github.ref_type == 'tag' && startsWith(github.ref_name, 'v') }}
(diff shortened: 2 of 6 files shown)


build.json
```diff
@@ -3,6 +3,9 @@
     "PYTHON_VERSION": {
       "default": "3.9.13"
     },
+    "PGSQL_VERSION": {
+      "default": "15"
```
unkcpz, line 7: Do we need also to pin the minor version? We can intentionally make more restrictions to avoid confliction.

unkcpz (commented):
Thanks! @danielhollas Looks all good, just one request to discuss. 
Can you try to retrigger the CI, build and test are not running?

build.json
yakutovicha: Postgres guarantees that API won't change for the same major version. I am confident that fixing the major version is good enough.

yakutovicha:
What happens to the tests? Does anyone know?

The changes look good to me. It would be great if the tests were also passing.

danielhollas:
I'll reteigger the tests. I think sometimes GHA get stuck for now reason
and doesn't run the jobs.

Thanks for the review!

Dne čt 2. 2. 2023 11:12 dop. uživatel Aliaksandr Yakutovich <
***@***.***> napsal:

danielhollas:
@yakutovicha @unkcpz tests are passing so this should be good to go.

yakutovicha (approved):

PR #359, opened 2023-02-16
Add BASH completion functionality

danielhollas:
As an alternative, we could install `bash-completion` package via APT, but using `mamba-bash-completion` from conda-forge makes it work for conda/mamba commands as well so I think it is preferable.

Closes #358 

Tested locally (image `ghcr.io/aiidalab/full-stack:pr-359`)

PR #360, opened 2023-02-16
Move ssh agent start script to home prepare script

### LINKED ISSUE #357: singlesshagent section in .bashrc is duplicated upon each container startup
@unkcpz I've noticen an issue with this code, the '.bashrc' gets appended everytime the container is started, so this code gets needlessly duplicated. I would proposed to move this in the `10_prepare-home-rc.sh`, where you can append it just one when the `.bashrc` is first copied over.

_Originally posted by @danielhollas in https://github.com/aiidalab/aiidalab-docker-stack/pull/353#discussion_r1096575554_
            
Would be good to solve before the next docker stack release.

unkcpz:
fixes #357 

The sshagent source command is added everytime where the notebook start. I move the append to bashrc command to the script which create the bashrc and only attach the command the first time bashrc created.


unkcpz:
Tested with aiidalab-launch locally and the `.bashrc` file looks all good.

unkcpz:
> Have you tested with a brand new environment (e.g. by running aiidalab-launch reset)

I test by creating a new profile. 

> Because in this case the .bashrc file already exists,

I think this is the expected behavior, no? The `.bashrc` is created only the first time.

unkcpz:
If you look at the earliest implement which the append is in the dockerfile https://github.com/aiidalab/aiidalab-docker-stack/blob/24787bf74403b4d5c1553eaa2e5e9c2fe3a28357/stack/base/Dockerfile#L48-L53. It has the problem that if the .bashrc exist in persistent volume, when the container instance is restarted (The current issue is when notebook restart), it will duplicate append the command.

danielhollas:
For me the bashrc is always there, even on the first startup so in the current implementation it never get appended.

Are you using bind mount or volume mount?

unkcpz:
The same problem as #359, I forget to update the image in the profile.

unkcpz:
I remember there is a discrepancy between the docker and deployment by using jupyterhub. The `.bashrc` is created if run using aiidalab-launch by docker. It needs a clever way to append the text.

stack/base/before-notebook.d/10_prepare-home-config.sh
```diff
@@ -0,0 +1,50 @@
...
+# If the container is start by spawner and the home is remounted.
+# The .bashrc in HOME won't be set properly.
+if [[ ! -f  /home/${NB_USER}/.bashrc ]]; then
+    cp /etc/skel/.bashrc /home/${NB_USER}/.bashrc
+fi
+
+# Set sshagent by source load-singlesshagent.sh script
+# append the command text of source to .bashrc if the script /opt/bin/load-singlesshagen.sh is present
+# and the command text is not already present in .bashrc
+if [[ -f /opt/bin/load-singlesshagent.sh ]] && ! grep -q "# Load singlesshagent on shell startup." /home/${NB_USER}/.bashrc; then
```
danielhollas, line 12: I would use a variable here

```
header="# Load singlesshagent on shell startup."
```

Otherwise looks okay, LMK when you want me to test this again @unkcpz

unkcpz: Yes, done!

danielhollas (approved):
LGTM! :rocket:  I tested locally and it works.

PR #362, opened 2023-02-24
Bump aiida-core to v2.2.2

danielhollas:
Version 2.2.2 contains a critical bugfix for the AiiDA daemon. 

https://github.com/aiidateam/aiida-core/blob/main/CHANGELOG.md#v222---2023-02-10

We should include this in the next release (together with the aiidalab updates.

### DIFF
--- a/build.json
+++ b/build.json
@@ -7,7 +7,7 @@
       "default": "15"
     },
     "AIIDA_VERSION": {
-      "default": "2.2.1"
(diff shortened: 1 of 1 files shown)


unkcpz (approved):
Thanks @danielhollas, I think this is a very good example of which aiida-core version we need to follow here for the docker stack.

Issue #363, opened 2023-02-24
Publish new version of the old docker stack with updated aiidalab and aiidalab-home


danielhollas:
We should probably create a new support branch for the old docker stack from the last commit before the transition.

danielhollas:
@unkcpz is this one done?

unkcpz:
@yakutovicha mentioned there is some problem that the PostgreSQL version is not correct, and he will take care of it today or tomorrow.

yakutovicha:
It is okay - I will take care of that in another repository. I think the issue can be closed.

PR #366, opened 2023-03-17
Bump aiidalab and aiidalab-home to v23.03.0

danielhollas:


danielhollas:
@yakutovicha @unkcpz Since there have been _a lot_ of (awesome :muscle:) changes, could we please give this a final test?
I'll then release a new version of the new Docker stack.

@unkcpz could you take care of releasing a new version of the old docker stack? (#363) Also please release AWB 1.4.3

unkcpz:
> @unkcpz could you take care of releasing a new version of the old docker stack? (https://github.com/aiidalab/aiidalab-docker-stack/issues/363) Also please release AWB 1.4.3

Sure, I'll do it. I think that need to wait until we finalize the aiidalab-widgets-base, and collect all changes (if there are more) to the 1.x support. The old stack has widget-base pinned. Can we just remove the version pinned for AWB, we have the aiidalab now will not show the incompatible AWB, which means the 1.4.x will be the valid version for the old stack. I'll take care of it anyway.

danielhollas:
>  The old stack has widget-base pinned.

I don't think this is true? Why would be AWB pinned in the old stack?

danielhollas:
I've give this a quick test and things seems to work, but a recent plumpy change uncovered a bug in my workchain, due to 
https://github.com/aiidateam/aiida-core/pull/5924

@unkcpz you might want to really test QeApp with this to verify the workchains work.

unkcpz:
I gave it a test by running calculations on QeApp, all works fine in the test. 
The issue opened in https://github.com/aiidalab/aiidalab/issues/378 was not related to the docker image and require more discussion. 
From my side, this is ready to merge.

unkcpz:
> > The old stack has widget-base pinned.
> I don't think this is true? Why would be AWB pinned in the old stack?

It is, see https://github.com/aiidalab/aiidalab-docker-stack/blob/support/aiida-1.x/requirements.txt#L2
I don't know why it is pinned, any idea @yakutovicha? Can I remove it or I wait for the AWB version to be released?

unkcpz (approved):

danielhollas:
>  I don't know why it is pinned, any idea @yakutovicha? Can I remove it or I wait for the AWB version to be released?

See discussion in #291, looks like it was a rather unwanted side-effect. But in any case, it is pinned to `~1.3` i.e. `>=1.3, <2.0` so you don't need to remove the constraint. So I think you're good to go w.r.t releasing AWB 1.4.3 and preparing the the old docker stack image, it needs to be released before AWB 2.0 anyway.

I've merge this PR, but waiting for @yakutovicha to confirm before making a new release (of the new docker stack).

yakutovicha:
> I've merge this PR, but waiting for @yakutovicha to confirm before making a new release (of the new docker stack).

it is hard to be against a new release. Go ahead 👍

PR #370, opened 2023-04-13
test: install awb in lab stack

danielhollas:
This is just a test to see how much would installing AWB into the image increased its size.


danielhollas:
Looks like the size is increased by 700Mb, which is quite significant. But perhaps this could be decreased if we tried to trim the AWB dependencies a bit, as per https://github.com/aiidalab/aiidalab-widgets-base/issues/392

PR #372, opened 2023-04-23
Arm64 support

unkcpz:
- [x] can only run specific stack build by: `doit build --target base --target full-stack`
- [x] Test using self-hosted arm64 runner. Hosted in AWS.
- [x] `mamba-bash-completion` package for arm arch


.github/workflows/ci.yml
```diff
@@ -109,8 +114,8 @@ jobs:
 
     integration-test:
         needs: build
-        runs-on: ubuntu-latest
-        timeout-minutes: 10
+        runs-on: [self-hosted, spot-self-hosted, ubuntu-latest]
```
unkcpz, line 117:
```suggestion
        runs-on: [self-hosted, ubuntu-latest]
```

.github/workflows/ci.yml
```diff
@@ -77,18 +78,22 @@ jobs:
...
+                  echo "targets=$targets" >> "${GITHUB_OUTPUT}"
+                  images=$(echo $(cat bake-meta.json | jq -c '. | [to_entries[] | {"key": (.key|ascii_upcase + "_IMAGE"), "value": .value}] | from_entries'))
+                  echo "images=$images" >> "${GITHUB_OUTPUT}"
 
     test:
         needs: build
 
-        runs-on: ubuntu-latest
-        timeout-minutes: 10
+        runs-on: [self-hosted, spot-self-hosted, ubuntu-latest]
```
unkcpz, line 95:
```suggestion
        runs-on: [self-hosted, ubuntu-latest]
```

stack/base-with-services/Dockerfile
```diff
@@ -8,18 +8,51 @@ WORKDIR /opt/
...
 RUN mamba create -p /opt/conda/envs/aiida-core-services --yes \
-     aiida-core.services=${AIIDA_VERSION} \
      postgresql=${PGSQL_VERSION} \
-     rabbitmq-server=3.8.14 \
      && mamba clean --all -f -y && \
      fix-permissions "${CONDA_DIR}" && \
      fix-permissions "/home/${NB_USER}"
 
+# Install RabbitMQ in a dedicated conda environment.
+# If  the architecture is arm64, we install the default version of rabbitmq provided by Ubuntu.
```
unkcpz, line 20:
```suggestion
# If  the architecture is arm64, we install the default version of rabbitmq provided by the generic binary, the version needs to be compatible with system erlang version.
```

tmp.json
```diff
@@ -0,0 +1 @@
+{"BASE_IMAGE":"ghcr.io/aiidalab/base:sha-df0a6ea","BASE_WITH_SERVICES_IMAGE":"ghcr.io/aiidalab/base-with-services:sha-df0a6ea","LAB_IMAGE":"ghcr.io/aiidalab/lab:sha-df0a6ea","FULL_STACK_IMAGE":"ghcr.io/aiidalab/full-stack:sha-df0a6ea"}
```
unkcpz, line 1: Remove

bake-meta.json
unkcpz, line 1: Remove the test file.

.github/workflows/ci.yml
```diff
@@ -151,45 +177,81 @@ jobs:
...
-                  password: ${{ secrets.DOCKER_PASSWORD }}
-                  repository: aiidalab/${{ matrix.target }}
-                  short-description: ${{ github.event.repository.description }}
-
-            - uses: softprops/action-gh-release@v0.1.14
-              name: Create release
-              if: startsWith(github.ref, 'refs/tags/v')
-              with:
-                  generate_release_notes: true
+    # release:
```
unkcpz, line 194: Back.

requirements.txt
unkcpz, line 1: Not used, remove.

unkcpz:
Hi @cpignedoli @mbercx, I think you have M1/M2 chip macbook that can test the image. 
To test it, you can pull the container by `docker pull ghcr.io/aiidalab/full-stack:pr-372`. Since it is the private image registry, you need to configure and authorize with your personal token, as instructed at https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-with-a-personal-access-token-classic

After pulling, you can set a new aiidalab-launch profile and chaning the image to `ghcr.io/aiidalab/full-stack:pr-372` and start the container. 

I test on my macbook and QeApp can be installed, the arch now shows `arm64` as expected. For the QeApp, the problem is that the conda forge for qe installation not supported yet, but I also made progress on having it https://github.com/conda-forge/qe-feedstock/pull/39

mbercx:
Thanks @unkcpz! I was trying to launch the container with `aiidalab-launch`, but I ran into an error:

```console
❯ aiidalab-launch profiles show test-pr372
port = 8889
default_apps = []
system_user = "mbercx"
image = "ghcr.io/aiidalab/full-stack:pr-372"
home_mount = "aiidalab_test-pr372_home"
extra_mounts = []
❯ aiidalab-launch start
Downloading latest version of 'ghcr.io/aiidalab/full-stack:pr-372'... done.
Starting container...
Error: Startup failed due to an unexpected error.
```

Unfortunately, there is no `--verbose` option to get more information (also see https://github.com/aiidalab/aiidalab-launch/issues/176):

```console
❯ aiidalab-launch start --help
Usage: aiidalab-launch start [OPTIONS]

  Start an AiiDAlab instance on this host.

Options:
  --restart                       Restart the container in case that it is
                                  already running.
  --wait INTEGER                  Time to wait after startup until all
                                  services are up. Set to zero to not wait at
                                  all and immediately return.  [default: 120]
  --pull / --no-pull              Specify whether to pull the configured image
                                  prior to the first start of the container.
                                  [default: pull]
  --no-browser                    Do not open AiiDAlab in the browser after
                                  startup. (This is disabled by default if the
                                  wait time is set to zero.)
  --show-ssh-port-forwarding-help
                                  Show guidance on SSH port forwarding.
  -f, --force                     Ignore any warnings and start anyways.
  -p, --profile TEXT              Select profile to use.
  --help                          Show this message and exit.
```

I tried running the image manually through docker, but it's been a while again since I've used any of this. Can we maybe debug via a Zoom call?

mbercx:
I seem to run into permission errors. Originally I had changed my `system_user`, but even with the default I'm getting:

```console
❯ aiidalab-launch logs
Entered start.sh with args: jupyter notebook --NotebookApp.default_url='/apps/apps/home/start.ipynb' --ContentsManager.allow_hidden=True
WARNING: no write access to /home/jovyan. Try starting the container with group 'users' (100), e.g. using "--group-add=users".
/usr/local/bin/start.sh: running hooks in /usr/local/bin/before-notebook.d as uid / gid: 1000 / 100
/usr/local/bin/start.sh: running script /usr/local/bin/before-notebook.d/10_prepare-home-config.sh
cp: cannot create regular file '/home/jovyan/.bashrc': Permission denied
Entered start.sh with args: jupyter notebook --NotebookApp.default_url='/apps/apps/home/start.ipynb' --ContentsManager.allow_hidden=True
WARNING: no write access to /home/jovyan. Try starting the container with group 'users' (100), e.g. using "--group-add=users".
/usr/local/bin/start.sh: running hooks in /usr/local/bin/before-notebook.d as uid / gid: 1000 / 100
/usr/local/bin/start.sh: running script /usr/local/bin/before-notebook.d/10_prepare-home-config.sh
cp: cannot create regular file '/home/jovyan/.bashrc': Permission denied
Entered start.sh with args: jupyter notebook --NotebookApp.default_url='/apps/apps/home/start.ipynb' --ContentsManager.allow_hidden=True
WARNING: no write access to /home/jovyan. Try starting the container with group 'users' (100), e.g. using "--group-add=users".
/usr/local/bin/start.sh: running hooks in /usr/local/bin/before-notebook.d as uid / gid: 1000 / 100
/usr/local/bin/start.sh: running script /usr/local/bin/before-notebook.d/10_prepare-home-config.sh
cp: cannot create regular file '/home/jovyan/.bashrc': Permission denied
```

unkcpz:
Hi @mbercx, do you need to run your docker with sudo? Maybe you install docker using something like the non-root mode? I recommend to use the classical installation and post-setting as https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user

danielhollas:
@mbercx just FYI in your aiidalab-launch profile the user **must** be `jovyan`. If you were using a different, I would recommend to start fresh by resetting the profile with `aiidalab-launch reset -p your_profile` (**WARNING** this will delete all associated docker volumes or mounts). btw: Are you using Docker volumes or bind mounts?

unkcpz:
I think @danielhollas is a correct, nice catch. I am thinking if we can not solve the displayed user modification issue, we remove it from aiidalab-launch profile for the moment?

danielhollas:
> we remove it from aiidalab-launch profile for the moment?

Yeah, that's not a bad idea, but the problem is backwards compatibility. Right now aiidalab-launch works for both old and new stack, and we should keep it that way at least for a while longer.

mbercx:
> @mbercx just FYI in your aiidalab-launch profile the user **must** be `jovyan`. If you were using a different, I would recommend to start fresh by resetting the profile with `aiidalab-launch reset -p your_profile` (**WARNING** this will delete all associated docker volumes or mounts). btw: Are you using Docker volumes or bind mounts?

Golden advice as always, @danielhollas. I had already considered adapting the user back to `jovyan`, but it seems resetting the profile was indeed what was needed.

Already managed to start the image, now testing the QEapp a bit. Will get back to you!

mbercx:
Alas, conda woes:

qeapp-condafailed

That progress bar is also really not happy about it. 😅

unkcpz:
Hi Marnik, thanks for the test. The problem is totally as expected. The qe conda-forge didn't have arm64 supported yet, the PR is open and I think it is very closed to be merged https://github.com/conda-forge/qe-feedstock/pull/39.

mbercx:
> Wait... the heart emoji is from some guy has the same github id as you 🤔 What a small world.

I have not been sleeping enough, and my memory is clearly starting to fail me. I will go lie down now and contemplate my life decisions. 😅

unkcpz:
> Yeah, that's not a bad idea, but the problem is backwards compatibility. Right now aiidalab-launch works for both old and new stack, and we should keep it that way at least for a while longer.

I think just from the point of usability, it is more easy to use if we fix the user and provide an additional option for `aiidalab-launch profiles add <profile-name>` to have the template for the old stack which have user `aiida` and have default image set to `aiidalab/aiidalab-docker-stack:22.8.1`.

unkcpz:
@danielhollas , I open the issue at https://github.com/aiidalab/aiidalab-launch/issues/177

stack/base/Dockerfile
```diff
@@ -17,10 +17,11 @@ WORKDIR /opt/
 ARG AIIDA_VERSION
 
 # Install the shared requirements.
+# XXX: bring back mamba-bash-completion once it is available on conda-forge.
+# https://github.com/conda-forge/mamba-bash-completion-feedstock/pull/2
```
unkcpz, line 21: @danielhollas, here is the workaround required for the `mamba-bash-completion` which not supported in arm64.

danielhollas (changes requested):
@unkcpz thanks for all your thankless work on this! I've tested it on a M2 Mac machine and it works great.

I mostly just have some clarifying questions, nothing big. I would only suggest we wait till the mamba-bash-completion package becomes available, which should hopefully not be long.

Just as a side note, I found out that I won't be able to run my app on Arm. My app relies on ORCA for ab initio calculations. While ORCA team provides binaries for Mac Arm64, what we actually need here are Linux/arm64 binaries so that they can run inside our Ubuntu container. I didn't quite realized this before. It is a bit unfortunate, I'll see if the ORCA folks would be willing to release a new binary (though our use case seems to be quite niche).

.github/workflows/ci.yml
```diff
@@ -68,7 +68,9 @@ jobs:
                     | tee docker-bake-template-meta.json
             - name: Build images with buildx bake
               id: bake
-              uses: docker/bake-action@v2.3.0
+              env:
+                  BUILDX_NO_DEFAULT_ATTESTATIONS: 1
```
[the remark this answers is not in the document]
unkcpz: This is not work actually, for the `docker/bake-action` the correct way to disable [attestation](https://docs.docker.com/build/attestations/slsa-provenance/) is as shown in https://github.com/docker/bake-action#inputs.
The attestation need to be disabled otherwise there will be an `unknown/unknown` image in the packages registry https://github.com/aiidalab/aiidalab-docker-stack/pkgs/container/base-with-services.

environment.yml
```diff
@@ -3,12 +3,12 @@ name: aiidalab-docker-stack
...
     - conda-forge
 dependencies:
-    - docker-compose=1.29.2
     - doit=0.36.0
     - dunamai=1.13.0
     - pip=22.2.2
     - pytest=7.1.2
     - requests==2.28.1
     - pip:
+          - docker-compose==1.29.2
```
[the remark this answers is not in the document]
unkcpz: Yes, the package not available for maxos-arm64.

tests/conftest.py
```diff
@@ -16,7 +16,7 @@ def is_responsive(url):
         return False
 
 
-@pytest.fixture(scope="session", params=["lab", "base-with-services"])
+@pytest.fixture(scope="session", params=["lab", "base-with-services", "full-stack"])
```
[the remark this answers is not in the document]
unkcpz: If you look at [stack/full-stack/Dockerfile](https://github.com/aiidalab/aiidalab-docker-stack/pull/372/files#diff-20a2c9d37a1aa2cd8ffb85dc4c0c8126e73ef093ae060c82563470a1dbdc8dac) you'll find I have re-install `erlang` even though it is already installed in the `base-with-services`. The reason is that we use the multiple stage and packages installed in the preliminary stage will not transfer to the final image. This test is will fail if `erlang` not installed explicitly in `full-stack`.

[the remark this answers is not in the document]
danielhollas: Right, fair enough. I guess my question then is whether there is enough reason to test the "lab" image separately, or whether just testing `base-with-services` and `full-stack` is enough (which would also allow us to get rid of bunch of extra skipping logic in some tests. But that can be left for another PR.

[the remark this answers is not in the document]
unkcpz: > whether just testing base-with-services and full-stack is enough

I am second on this. I didn't see the point to test `lab` and `full-stack`. Can you open an issue and assign it to me and you? Should be a quick fix I assume.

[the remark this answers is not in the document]
unkcpz: I tried to remove the test of `base-with-services` in https://github.com/aiidalab/aiidalab-docker-stack/pull/385. The lab is worth testing because we want to be sure strating with docker-compose is working.

danielhollas:
Hey @unkcpz are there any remaining blockers here in terms of all dependencies? Is the QeApp actually running on this Arm image? If yes, maybe @mbercx and @cpignedoli can do final round of testing and we're good to go?

unkcpz:
> Is the QeApp actually running on this Arm image?

Is not, but nothing can be done here. What I need to do is update the qe installed from conda-forge to 7.2 which starts to have arm64 supported. 

>  If yes, maybe @mbercx and @cpignedoli can do final round of testing and we're good to go?

Maybe @cpignedoli can test on Empa apps? I have apple M1 laptop as well so I can give it a test for QeApp. I plan to release a new version with the "daemon stop" fix. Let's see how fast I can make a QE update for the app.

unkcpz:
Running a test on my M1 chip laptop with the QE from conda-forge, and all is good https://github.com/aiidalab/aiidalab-qe/pull/404 🎊

cpignedoli:
@unkcpz I will test everything hopefully by tonight.

cpignedoli:
I had no problems with my tests so from my side, despite I did not go through the files, I am OK with merging

unkcpz:
> I just noticed a minor bug: in the viewer, if I click on an atom it does not get selected, but if I enter the ID in the selection tab and click "apply selection" it works. Did you experience something similar?

Not in the amd64 machine. Is it the `nglviewer` that needs to respond? Can you check the version of `nglview`. 
But I'd like to tackle it in another PR. I want to merge this one and test https://github.com/aiidalab/aiidalab-docker-stack/pull/384, so there is no delay for the release. 

@cpignedoli could you open an issue for it with one screenshot? Thanks!

PR #373, opened 2023-04-23
Using sha as image reference in action

### LINKED ISSUE #328: Github Actions warnings
See for example here:

https://github.com/aiidalab/aiidalab-docker-stack/actions/runs/3539919228

image

To migrate the `ci.yml` see 

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files

> Starting today runner version 2.298.2 will begin to warn you if you use the save-state or set-output commands via stdout. We are monitoring telemetry for the usage of these commands and plan to fully disable them on 31st May 2023. Starting 1st June 2023 workflows using save-state or set-output commands via stdout will fail with an error.

So this needs to be done before the end of May. It may also affect other repositories.

unkcpz:
This is required by https://github.com/aiidalab/aiidalab-docker-stack/pull/372, the `BAKE_METADATA` was used to get the build targets (that is a list of `["base", "base-with-services", "lab", "full-stack"]`) and get the full image reference in ghcr.io.  
For instance, the images in the test and release workflow jobs are retrieved by using the `pr-<num>@sha:<unique sha>`. The sha is the digest of the image which generated in building and has to be get from `BAKE_METADATA`. 

The problem of https://github.com/aiidalab/aiidalab-docker-stack/pull/372 is that the `BAKE_METADATA` as an environment variable is too large therefore leading to "Argument too long" issue of `/bin/bash`. 

Instead, I add tags from PR commit `sha` for the image, and it can be read from `BAKE_TEMPLATE_METADATA` which is small.

This PR also include deprecating using `set-output` in github action.

fixes #328


unkcpz:
@danielhollas @superstar54 @yakutovicha anyone wants to have a look at this? Required by #372 to support arm64.

danielhollas:
@unkcpz this PR is a bit tricky for review since the YAML code is a mess (not your fault). I would suggest to split this PR into two, one with the `set-output` changes, and one with the SHA changes. That should hopefully make it clearer of what is going on.

unkcpz:
I did it separately and you can find the change of deprecate using set-output in the first commit https://github.com/aiidalab/aiidalab-docker-stack/pull/373/commits/1e41dbd7073b3659b234af4b33807e94c62a4584

Rebase and ready for review.

.github/workflows/ci.yml
```diff
@@ -77,11 +78,15 @@ jobs:
             - name: Set output variables
               id: bake_metadata
               env:
-                  BAKE_METADATA: ${{ steps.bake.outputs.metadata }}
```
danielhollas, line 80: Does this mean that the 'bake' step could be removed entirely?

danielhollas (commented):
Just wondering out loud. When I was looking into doing something similar in my own docker stack, one issue with using sha tag is that in that case each individual commit gets its tagged image, so the number of tagger images on ghcr.io will skyrocket unless they are being deleted / untagged somehow. I am not sure if that has any practival implications (e.g. I don't know the limits / capacity that we have on ghcr.io)

unkcpz:
> one issue with using sha tag is that in that case each individual commit gets its tagged image, so the number of tagger images on ghcr.io will skyrocket unless

But the previous solution is not helpful for this, new images will created and pushed for every commits. Using `sha` is just an extra tag for the `pr` tag, they share the same image, if you go to package page you'll find:

image

So I think it is safe at least compared to previous usage.

It also has the advantage that the image is easier to distinguish by commit sha. Otherwise, you need to use the long digest if you want to point to the image generated by the previous commit after you make new commit.

.github/workflows/ci.yml
unkcpz: The step `id` can be removed, but the baking step is the actual step for building the image.

danielhollas:
> `Using sha is just an extra tag for the pr tag, they share the same image`

That's not exactly true though, because if you have more than one commit in a PR, that all of them will be tagged, versus previously only the latest one would be tagged with `pr-XXX`. 

The changes themselves look ok, but to have full confidence I would have to look deeper to verify that everything is passed around correctly. Since this is a potentially high-impact change if we didn't get this right, @yakutovicha can you please also take a look?

unkcpz:
> That's not exactly true though, because if you have more than one commit in a PR, all of them will be tagged, versus previously only the latest one would be tagged with pr-XXX.

But compared to what it had before, every time there are new commits, the build will produce new images and use that for test. Instead of that every commit will trigger the CI but only the last commit corresponds to the CI, it only builds for the final commit which is guaranteed by GitHub.

.github/workflows/ci.yml
```diff
@@ -77,12 +78,16 @@ jobs:
...
+                  BAKE_TEMPLATE_META: ${{ steps.meta.outputs.bake-file }}
               run: |
-                  targets=$(echo "${BAKE_METADATA}"| jq -c 'keys')
-                  echo "::set-output name=targets::${targets}"
-                  images=$(echo "${BAKE_METADATA}"| jq -c '. as $base |[to_entries[] |{"key": (.key|ascii_upcase|sub("-"; "_"; "g") + "_IMAGE"), "value": [(.value."image.name"|split(",")[0]),.value."containerimage.digest"]|join("@")}] |from_entries')
-                  echo "::set-output name=images::${images}"
+                  cat ${BAKE_TEMPLATE_META} | jq -c \
+                    | .github/workflows/merge-bake-template-target.sh \
```
danielhollas, line 84: It looks like these lines are the same as on lines 66-67 in "Generate docker-bake meta file".
So instead you could reuse it and read the `docker-bake-template-meta.json` file directly here to avoid duplication?

danielhollas (changes requested):
Still not entirely convinced that this is the best solution, but we can try it out. Just one small suggestion.

.github/workflows/ci.yml
unkcpz: Changed and tests passed as usual :wink:

danielhollas (changes requested):
Thanks @unkcpz.
I have taken a closer look at the outputs from the CI runs. While we're here, let's get rid of the rest of the GHA warnings.

1. Update actions/checkout to v3 to get rid of the NodeJs warnings
image

2. Update `docker/bake-action` to get rid of `save-state` warnings. Looks like bumping to v2.3.0 should do the trick, according to https://github.com/docker/bake-action/releases and https://github.com/docker/setup-buildx-action/issues/168

image

More importantly though, I think you also need to test the Release job, it is not clear to me whether the changes here will not break it. I would recommend temporarily modifying it so it can run on this PR, but commenting out the actual publishing steps, i.e. everything below and including  "Push image" step.

danielhollas:
@unkcpz looks like some other warnings are coming from `provision-with-micromamba`, updating to v14 should fix it.
Also please get rid of the deprecated `set-output` in the Release job.

unkcpz:
All the env variables from `build` is shown in https://github.com/aiidalab/aiidalab-docker-stack/actions/runs/4880938054?pr=373 of release actions as expected.

.github/workflows/ci.yml
```diff
@@ -193,3 +195,12 @@ jobs:
               if: startsWith(github.ref, 'refs/tags/v')
               with:
                   generate_release_notes: true
+            - name: test push image
```
danielhollas, line 198: @unkcpz this should be removed right? Otherwise looks good.

unkcpz: Correct! Purged.

danielhollas (approved):

PR #374, opened 2023-05-02
Run for specific target

unkcpz:
This allows to run build for a specific stack by `doit build --target base --target base-with-services`. Otherwise, all stacks are builded and takes time during developing.


danielhollas:
> The installation failed everywhere because of pybtex which I guess from pymatgen as well. But it seems they pinned it to 0.24 that is a version should work. Inside the container it tried to install pybtex=0.22 and cause the problem.

Not sure I understand. Is `pybtex` dependency coming from `pymatgen`? So this needs a fix in the QeApp and not in the Docker image?

dodo.py
```diff
@@ -10,9 +12,18 @@
 
 DOIT_CONFIG = {"default_tasks": ["build"]}
 
+_ARCH_PLATFORM_MAPPING = {
+    "aarch64": "linux/arm64",
```
danielhollas, line 16: Should this be in this PR? Or does it belong to #372?

danielhollas (commented):
I've never used `doit` or bake for local development so I am not the best for review. It looks fine at a glance, but I do not understand whether this PR depends on #372 or not.

dodo.py
unkcpz: This has nothing to do with #372, but running with `--platform arm64` will fail since #372 has not yet applied. 
As you can see from the initial code, it tries to support aarch64 by mapping `aarch64` to `linux/arm64`, it is still valid if run from an amd64 machine and the platform will get from docker info and set to be amd64.

danielhollas (approved):

unkcpz:
🚀

PR #376, opened 2023-05-08
Integration test for stable AWB and QEApp

unkcpz:
We now have AWB 2.0 released. The integration test can be run.


danielhollas (approved):
🎉

danielhollas:
It might be because we don't uninstall the AWB version from the previous test.

unkcpz:
> It might be because we don't uninstall the AWB version from the previous test.

Correct, now only aiidalab-qe integration test not passed, I added the backport to branch `support-v23.4.x` and make the release `v23.4.2`. It needs to be picked as stable by https://github.com/aiidalab/aiidalab-registry/pull/122. 
@danielhollas @yakutovicha please have a look when you get time. No hurry, we are waiting for aiida-core `v2.3.1` release.

tests/test_aiidalab.py
```diff
@@ -130,12 +130,16 @@ def test_install_apps_from_stable(aiidalab_exec, package_name, nb_user, variant)
     assert "dependency conflict" not in output
     assert f"Installed '{package_name}' version" in output
 
+    # Uninstall the package to make sure the test is repeatable
+    aiidalab_exec(f"aiidalab uninstall --yes --force {package_name}", user=nb_user)
```
danielhollas, line 134: I think we need to implement this as a separate fixture (`aiidalab install`?) and have this as a teardown, not as part of the test itself, otherwise things will still be flaky...

danielhollas (changes requested):

tests/conftest.py
```diff
@@ -87,3 +87,20 @@ def aiidalab_version(_build_config):
...
+def generate_aiidalab_install_output(aiidalab_exec, nb_user):
+    def _generate_aiidalab_install_output(package_name):
+        output = (
+            aiidalab_exec(f"aiidalab install --yes {package_name}", user=nb_user)
+            .decode()
+            .strip()
+        )
+
+        # Uninstall the package to make sure the test is repeatable
+        aiidalab_exec(f"aiidalab uninstall --yes --force {package_name}", user=nb_user)
```
danielhollas, line 102: Hmm, I don't think this works when you pass in `package_name` as the whole url to github. Just tested this locally

```console
$ aiidalab uninstall aiidalab-widgets-base@git+https://github.com/aiidalab/aiidalab-widgets-base
...
...
$ aiidalab uninstall aiidalab-widgets-base@git+https://github.com/aiidalab/aiidalab-widgets-base
Collecting apps to uninstall... Done.

Nothing to uninstall, exiting.
```

danielhollas (changes requested):

tests/conftest.py
danielhollas: Just had another idea. I think it would be good to run `pip check` right after the install to ensure there are no dependency conflicts. We kind of test it I think by inspecting the output, but this would be more robust I think.

unkcpz: Yes, `pip check` is a good idea, added. 

> , I don't think this works when you pass in package_name as the whole url to github. Just tested this locally

True, please check the new commit https://github.com/aiidalab/aiidalab-docker-stack/pull/376/commits/1d027eade531895b4cab598e8d93b05283a026a1, I extract the name of the app and uninstall it.

danielhollas (approved):

Issue #375, opened 2023-05-05
The demo images with QE installed


unkcpz:
Two images are good to have:
- base with QE and aiida-quantumespresso installed in advance. This is a image that can be used for QE tutorial or for quickly having an isolated aiida-quantumespresso development environment.
- demo image for QeApp. This is an image that has QeApp installed and ready to use out of the box. Currently, any app can be set as the default app of an AiiDAlab deployment but to start use the App the computer/codes are required to be setup. QeApp having the post-install script and the setup widget for installing QE from conda-forge and setup codes in localhost if it is not set yet. But it happened the first time QeApp startup which jeopardise the users' experience.

For qe from conda-forge, it can be installed in a separated image and copy to the target image using the multi-stage build strategy. It is what full-stack already used, it base on the `lab` image and having services from `base-with-services` image. We can having a dedicate image called `services` and copy things to `base-with-services` instead, to keep it alligment that having very similar structure as the `lab` and `full-stack`. This can also be the case for the qe conda forge.

unkcpz:
This is implemented in #402

PR #377, opened 2023-05-09
Decouple services stack

unkcpz:
As mentioned in #375, this is the attempts to move the services stack out and have `base-with-services` and `full-stack` almost the same. This will make it easy to add "qe" env more easily for both aiida and lab based stacks.


unkcpz:
This is not useful, it is not a good practice to have images only to store packages, I close it for the moment.

PR #380, opened 2023-05-10
Handle the aiida daemon PID stale exception

### LINKED ISSUE #379: Can not start the container if the aiida daemon is not properly stopped
Step to reproduce, using `full-stack:edge` (which has `aiida-core==2.3.0`, the tag is not exist since we did not release new stack yet.)
1. start the container with aiidalab-launch start -p <profile>
2. stop the container with aiidalab-launch stop -p <profile>
3. start again and it will fail.

The log shows:
```
++ verdi daemon stop
Profile: default
Stopping the daemon... FAILED
Critical: The daemon could not be reached, seemingly because of a stale PID file. Try starting the daemon to remove it and restore the daemon.
```
Exception happened at https://github.com/aiidalab/aiidalab-docker-stack/blob/5a1de2efdf86aab6ab2153eebf9cc728f31d9e68/stack/base/before-notebook.d/40_prepare-aiida.sh#L70-L73

I guess the problem comes from https://github.com/aiidateam/aiida-core/commit/8d963c98fbb3b1224cd69f3860f09360287034b3 where it originally will ignore the stale PID file and create a new one, but now it seems to exit with an exception. @sphuber can you confirm with this guess? 

I think a quick fix can be before running `verdi daemon stop` check and deleting the stale PID files manually.

unkcpz:
fix #379 
Require https://github.com/aiidateam/aiida-core/pull/6007 and with the new release of it.

### DIFF
--- a/stack/base/before-notebook.d/40_prepare-aiida.sh
+++ b/stack/base/before-notebook.d/40_prepare-aiida.sh
@@ -68,7 +68,7 @@ fi
 verdi profile show || echo "The default profile is not set."
 
 # Make sure that the daemon is not running, otherwise the migration will abort.
-verdi daemon stop
+verdi daemon clean_stale_pid_file && verdi daemon stop
 
 # Migration will run for the default profile.
 verdi storage migrate --force


danielhollas:
Thanks again @unkcpz for working on this. Based on the discussion in https://github.com/aiidateam/aiida-core/pull/6007, I am closing this PR, and instead we'll just upgrade to aiida-core 2.3.1 once it is available. Before that is done, we need to take care to NOT release a new docker-stack version.

unkcpz:
No problem, I forget to close this. Thanks for taking care of it. Cheers!

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #381, opened 2023-05-15
Implement pre-installed QE components for demo server

unkcpz:


danielhollas:
@unkcpz I am a bit wary of addind QeApp-specific code in this repo, because imo we should try to avoid getting QeApp a special treatment if AiiDAlab wants to be general-purpose platform. Wouldn't it be better to do it e.g. in https://github.com/aiidalab/aiidalab-docker-stack-demo

unkcpz:
I am not sure, just refer to how jupyter people maintain the docker stacks, they have all the customized images (`scipy`, `tenserflow`, ...) in the same repo https://github.com/jupyter/docker-stacks and build from the base image.
I can see with our current github action implementation, that maybe tucking everything in one build action job is not ideal. 

I also find put it inside the same repo maybe make us easier to maintain just one repo without duplicating all the `doit`, `action` tools, etc. @mbercx, what do you think, it will be you and me working on this mostly. The goal is to provide ready-to-use image for aiida tutorial and for having a demo server for next review.

danielhollas:
> I am not sure, just refer to how jupyter people maintain the docker stacks, they have all the customized images (scipy, tenserflow, ...) in the same repo https://github.com/jupyter/docker-stacks and build from the base image.
I can see with our current github action implementation, that maybe tucking everything in one build action job is not ideal.
> I also find put it inside the same repo maybe make us easier to maintain just one repo without duplicating all the doit, action tools, etc. @mbercx, what do you think, it will be you and me working on this mostly. The goal is to provide ready-to-use image for aiida tutorial and for having a demo server for next review.

All fair points. This is of course all up to you. Just pointing out that the machinery that is used in this repo is not strictly necessary in case all you want is to extend the full-stack image. This is what I am doing for my app, with one single Dockerfile. (the github actions file is still annoyingly complex though).

https://github.com/ispg-group/aiidalab-ispg-docker-stack

btw: Wouldn't this QeApp image be a natural solution for deploying to the Quantum Mobile as discussed today? You wouldn't need to mess with `aiidalab-launch` at all since everything would be preinstalled.

unkcpz:
> btw: Wouldn't this QeApp image be a natural solution for deploying to the Quantum Mobile as discussed today? 

I think yes, but I'll probably try the nasty way to have a QuantumMobile image with QeApp, just make it work.

unkcpz:
Here it is! @mbercx you can continue adding `aiida-pseudo` after https://github.com/aiidateam/aiida-pseudo/pull/135 is merged.

To build the image locally, just create the environment by `conda env create -f environment.yml`, checkout to the branch, and run `doit build --target base --target qe`. 
With the dependencies pre-installed and QE installed, it is already superfast to have a workable QeApp right after starting the container.

mbercx:
> @mbercx, what do you think, it will be you and me working on this mostly.

I don't really feel qualified to have an opinion on this at the moment. I've only done a bit of tweaking of the stack here and there.

> Here it is! @mbercx you can continue adding aiida-pseudo after https://github.com/aiidateam/aiida-pseudo/pull/135 is merged.

Sure! Should find some time for that at some point, as well as finalizing that PR. 😅 Note that doing this would also add the `aiida-pseudo` archives to this repo, which may be another strike against adding QEapp-specific components to the `aiidalab-docker-stack`.

tests/test_qe.py
```diff
@@ -0,0 +1,67 @@
...
+    try:
+        os.mkdir(sdir)
+    except FileExistsError:
+        pass
+    return sdir
+
+
+def test_pw_executable_exist(aiidalab_exec, qe_version, variant):
+    """Test the rabbitmq-server can start, the output should be empty if
+    the command is successful."""
```
superstar54, line 58: Please update the docstring.

tests/test_qe.py
```diff
@@ -0,0 +1,67 @@
...
+        selenium.find_element(By.ID, "notebook-container")
+
+        return selenium
+
+    return _selenium_driver
+
+
+@pytest.fixture
+def final_screenshot(request, screenshot_dir, selenium):
+    """Take screenshot at the end of the test.
```
superstar54, line 36: This is not used in the test.

superstar54 (changes requested):
Thanks for creating this image. 
- Could you add a test for aiidalab_qe module, e.g check can be imported or not?
- Should we already install the code and pseudo in the image so that the user will not be blocked in the first run?

unkcpz:
blocked by https://github.com/aiidalab/aiidalab-docker-stack/pull/390

The original CI makes adding new stack very hard to test.

unkcpz:
superseded by #402

PR #382, opened 2023-05-23
Bump aiida-core to 2.3.1 to fix daemon stop issue

unkcpz:


danielhollas:
@unkcpz will you take care of the conda package?

stack/base/Dockerfile
```diff
@@ -55,7 +55,7 @@ ENV AIIDA_USER_INSTITUTION Khedivial
 # https://aiida.readthedocs.io/projects/aiida-core/en/v2.0.0/howto/ssh.html#starting-the-ssh-agent
 # The startup of this script is configured in the before-notebook.d/20_setup-ssh.sh file.
 RUN wget --quiet --directory-prefix=/opt/bin/ \
-     "https://aiida.readthedocs.io/projects/aiida-core/en/v${AIIDA_VERSION}/_downloads/4265ec5a42c3a3dba586dd460c0db95e/load-singlesshagent.sh"
+     "https://aiida.readthedocs.io/projects/aiida-core/en/latest/_downloads/4265ec5a42c3a3dba586dd460c0db95e/load-singlesshagent.sh"
```
danielhollas, line 58: This is a very unfortunate dependency on the docs. The hash(?) in the URL also doesn't inspire confidence. Where does it come from and is it stable? Unfortunately, I have no idea how the whole docs / Sphinx system works.

@sphuber any thoughts if we could use a better URL here?

stack/base/requirements.txt
```diff
@@ -1,2 +1,2 @@
-aiida-core>=2.0.0,<3
+aiida-core>=2.0.0,<3,!=2.3.0
```
danielhollas, line 1: Nice! :clap:

stack/base/Dockerfile
danielhollas: @unkcpz FYU I've pushed this hotfix since the build was still failing. Take a look.

unkcpz: I don't think this file will be changed in the foreseeable future, so maybe just copy and paste it to the repo and maintain the copy here?

PR #383, opened 2023-05-23
Provide mlocate for running unit tests in container

unkcpz:
I think we'd better have this installed pre-hand, so users can run unit tests if they want even on the remote deployment, otherwise has to install it manually as root.

### DIFF
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -10,6 +10,7 @@ RUN apt-get update --yes && \
     # for apps which need to install pymatgen:
     # https://pymatgen.org/installation.html#installation-tips-for-optional-libraries
     build-essential && \
+    mlocate && \
     apt-get clean && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /opt/


danielhollas:
This is indeed very annoying, but I would actually suggest to go the other way and try to get rid of the `mlocate` dependency in tests. Do we know exactly where it comes from? Can we get rid of it? mlocate is I think something that I'd like to avoid in a production container, as it is a service that we shouldn't need. 

If we can't get rid of it from tests, maybe we could just have a dirty mlocate script that would produce the expected output?

danielhollas (commented):
Thanks, that was helpful. I looked at the code, and the thing is the the `locate` command is used by `pgtest` only as a fallback when other things fail. What the underlying problem is that we install psql in a separate conda environment `aiida-core-services` and so the `pg_ctl` command is not in PATH. However, I just tested that when I activate this environment with `conda activate aiida-core-services` I can run the tests. @unkcpz can you give it a try.

If that works for you than honestly I would not do anything here and just accept that you need to activate the environment. Alternatively, we could perhaps make a symlink of `pg_ctl` to some folder that's in PATH or some other hack, but thb that's just asking for problems in the future...

unkcpz:
> However, I just tested that when I activate this environment with conda activate aiida-core-services I can run the tests. @unkcpz can you give it a try.

That works! Thanks! 

I think we need somewhere for the developers of the app to know about this.

unkcpz:
I put the instruction in the readme of qeapp repo for the moment, https://github.com/aiidalab/aiidalab-qe/pull/403

danielhollas:
Cool, I've also opened an issue on the pgtest package to better handle this case

https://github.com/jamesnunn/pgtest/issues/25

PR #384, opened 2023-05-23
Upgrade pip to v23.0.1

danielhollas:
Pip changelog:
https://pip.pypa.io/en/stable/news/#v23-0-1


unkcpz:
Thanks @danielhollas! I have a vague memory that there is some reason the pip version needs to be pinned under `<22.3`, related to some setuptool issue. I'll have a search. If everything is just fine, sure, we need to use the latest version.

danielhollas:
Good call, thanks! It looks like the pip version was constrained here https://github.com/aiidalab/aiidalab-docker-stack/pull/271, but the the underlying issue was a deprecation of a pip cmdline option that we've been using in `aiidalab` package. But was fixed a long time since https://github.com/aiidalab/aiidalab/pull/297 so I think we should be good here. But I will test this thoroughly.

unkcpz:
Thanks @danielhollas. I rebase the arm64 support PR, for safety let's get that merge first and have this PR test for the arm64 (I think it will be all fine, the packages from pip always noarch specific).

unkcpz (approved):
All good! thanks @danielhollas

PR #389, opened 2023-06-27
Test Python 3.11 image :eyes: 

danielhollas:


danielhollas:
The image builds, but AWB cannot be installed, since `optimade==0.18` is not available for Python 3.11. 
This dependency needs to be unpinned / upgraded in the `optimade-client` package (to at least version `optimade==0.21.0`.

https://github.com/CasperWA/voila-optimade-client/blob/5f4ffbbe66aed1406e09601ad8af7fe241d6bd2a/requirements/requirements_base.txt#L6

danielhollas:
I've opened an issue on the `optimade-client` package.

https://github.com/CasperWA/voila-optimade-client/issues/506

danielhollas:
Just did a quick test, and it looks like upgrade to Python 3.11 would speed up loading of my application by ~600ms. :tada: 

So definitely worth to pursue this, but obviously we first need to update to 3.10.

danielhollas:
Closing for now, need to upgrade to 3.10 first.

danielhollas:
Just a note for my future self, the version docker.io/jupyter/minimal-notebook:python-3.11.3 is the last version that ships with notebook==6.5 so that's what we should target.

PR #390, opened 2023-06-28
Using reusable CI workflows for docker stack build

unkcpz:
This is a rearrangement of docker build CI. The CI was run to build everything (every target with two architecture, that is 8 things to build) in one action which makes the CI takes very long and hard to debug. The tests also was happened within one action for all the docker stacks. 

I adopt the way how jupyter community managing their CI test, https://github.com/jupyter/docker-stacks. The idea is to using github reusable workflow without duplicating the CI workflows. 

The final outcome of build didn't change too much, for every new PR, once it successfully build and tests passed, the docker image with tag `pr-xxx` is uploaded to gitub container registry https://github.com/orgs/aiidalab/packages. https://github.com/aiidalab/aiidalab-docker-stack/pkgs/container/full-stack/105900232?tag=pr-390 is the full-stack image by this PR. For different commit of the PR, it is distinguished by new `sha-xxx` tag. The `pr-xx` is pointed to the latest built images(and tests passed, this is different from the previous implementation where the unit tests are run after upload to `ghcr`, it is easy to separate the tests from the build and make it independent of upload to ghcr, @danielhollas @yakutovicha any opinion on this? I prefer only to upload the test-passed images).

A new thing is for every image build, the image is exported to the workflow artifacts and kept there for 3 days, this can be downloaded and tested locally. The artifacts will be uploaded even if the unit tests are failed. The artifacts of CI workflow by this PR is in https://github.com/aiidalab/aiidalab-docker-stack/actions/runs/5421607053?pr=390

Moreover, the CI workflow is clear and nice IMO:

image

- [x] clear name between "arch" and "platform", the arch is either "x86_64" and "aarch64", the platform is what used in docker build, is either "linux/arm64" or "linux/amd64".


unkcpz:
The push to dockerhub will only happened when making the release (when `github.ref` is `tags/v*`) but I trigger it manually and test on my own dockerhub repo https://hub.docker.com/repository/docker/jusong/base/tags?page=1&ordering=last_updated

unkcpz:
Since it is hard to review this one and the PR is needed for further developments, I made the decision to merge and release to test. If anything breaks on dockerhub I can always push new to override. Some actions can only be tangibly tested when released.

docker-bake.hcl
```diff
@@ -40,10 +45,8 @@ variable "TARGETS" {
 function "tags" {
   params = [image]
   result = [
-    "${REGISTRY}${ORGANIZATION}/${image}:${VERSION}",
-    "${REGISTRY}${ORGANIZATION}/${image}:python-${PYTHON_VERSION}",
```
danielhollas, line 44: @unkcpz why was this removed? I think we want to keep using these tags on Dockerhub, no? Also, what is the purpose of the `newly-build` tag?

btw: Thanks for working on this! Looks much cleaner now.

unkcpz: This is not used for push to the remote registry. When push to the remote, tags are defined from action it self, see [here](https://github.com/aiidalab/aiidalab-docker-stack/blob/5ed9b82f11ee02c1dac85d0afc00d810eba45e59/.github/workflows/docker-push.yml#L59-L66)

> Also, what is the purpose of the newly-build tag?

In the merge architecture action, the images are downloaded from artifacts uploaded. The newly-build tag is used to point to the image built from the github workflow. It can be `latest`, but if the latest is not found from the downloaded artifact, there are also remote `latest` tag that will be pulled automatically from the registry, which may cause the issue that the test may not run on the image that builds from the PR.

danielhollas: Thanks for the explanation. How are the latest and new-build tags distinguished when you have multiple PRs?

unkcpz: If you mean the images build in one PR, if a new commit is pushed to remote, I use concurrency workflow to make sure only one build is running https://github.com/aiidalab/aiidalab-docker-stack/blob/c97288e0c7b98cd6c52002f56aca6cc6a7e14310/.github/workflows/docker.yml#L55-L59

Thus the `pr-xxx` and `sha-xxx` tag will be generated only from one specific commit.

PR #393, opened 2023-07-04
Update 10_prepare-home-config.sh

ramirezfranciscof:
The `.ssh` folder may already exist before we run the initialization files (for example, if users need to set up keys via Dockerfile commands).


ramirezfranciscof:
Not the problem

PR #394, opened 2023-07-05
Configure Jupyter server to kill Python kernels after 12 hours

### LINKED ISSUE #378: Configure Jupyter server to kill idle python kernels
When the jupyter notebook server looses a connection to the frontend, it seems it will keep retrying indefinitely, which may lead to a memory leak as discussed in https://github.com/aiidalab/issues/issues/13. Moreover, it will leave the python kernel running indefinitely which needlessly consumes resources.

We need to configure the server to automatically kill inactive connections and kernels. Here are some resources on this:

https://discourse.jupyter.org/t/which-is-the-correct-way-to-cull-idle-kernels-and-notebook/8123
https://voila.readthedocs.io/en/latest/customize.html#cull-idle-kernels
https://github.com/jupyter/notebook/issues/4634

danielhollas:
When a Jupyter notebook server looses a connection to the frontend, it keeps the messages in a buffer. If there is a background thread running and trying to update the frontend, the buffer grows indefinitely, eventually consuming all available RAM. See https://github.com/aiidalab/issues/issues/13 for detailed analysis. 

Moreover, it will leave the python kernel running indefinitely which needlessly consumes resources.

We could turn this behaviour off by setting `buffer_offline_messages=False`,  but that may come with other problems for people with flaky internet connections. Instead, here we configure Jupyter to kill all kernels that have been alive for more than 12 hours. We also close all inactive terminals after 10 minutes.

Closes #378. Closes https://github.com/aiidalab/issues/issues/13


unkcpz (approved):
@danielhollas thanks for looking at this! Very important performance enhancement.

PR #396, opened 2023-07-10
Bump aiida-core to v2.4.0 and aiidalab to v23.3.2

unkcpz:
~~The workflow triggering paths are added and~~ bump the version of aiida-core to v2.4.0


danielhollas:
AiiDA 2.4.0 is very recent so based on passed experience maybe I would wait a bit before we switch (unless there is something that you need ASAP?). Also there's a database migration, with some potential manual steps, so we'll have to test carefully.

unkcpz:
Fair, I separate the changes to https://github.com/aiidalab/aiidalab-docker-stack/pull/397

unkcpz:
We are about to release aiida-core `v2.5.0` soon. So this one we can start to try. 
I also think maybe we can always follow the aiida-core minor version here in `main` branch. Because merge to `main` will influence the `edge` tag rather and only when making new release we decide whether to cherry-pick the version bump of `aiida-core`, what do you think? @danielhollas

danielhollas:
We still need to figure out / test how the DB migration is handled.

unkcpz:
> We still need to figure out / test how the DB migration is handled.

As I know, it handled by these command
https://github.com/aiidalab/aiidalab-docker-stack/blob/2d35cca2b8060b158121579ee4149b5d90fceadb/stack/base/before-notebook.d/40_prepare-aiida.sh#L73-L74

But it runs automatically without notifying the users, which might cause some issues, I think @yakutovicha had more expedience on this topic.

yakutovicha:
> But it runs automatically without notifying the users, which might cause some issues, I think @yakutovicha had more expedience on this topic.

This is precisely the migration line. What exactly is the issue?

unkcpz:
I can not foresee any immediate issues, when I wrote this I was think about maybe user do not what to migrate or what to trigger it manually with more control. I assume that was one of the idea to have a control app to do this?

danielhollas:
Well for one we need to test that it indeed works  before merging.

Also the disadvantage is that this migration is hidden from the user. In
this particular case, there is an extra message at the end of the migration
with a command to recompute hashes of the calculation nodes due to a
caching fix. We currently do not enable caching so it is not critical, but
it is problematic that the user is completely unaware of this, unless they
check the container startup logs.

For the future, it would indeed be great if migration was explicitly
triggered by the user I the control page imo.

Dne út 17. 10. 2023 10:01 uživatel Jusong Yu ***@***.***>
napsal:

yakutovicha:
Sure, but why does it block this specific PR? This migration has been there for a while and never given problems (unless I am unaware). And we did go through database migrations in the past.

To be practical, I would suggest merging the PR (provided that tests are successful). Before releasing a new version, I could test the new image on dev-aiidalab server. If it fails, we stop everything and we look closely at the issue.

In the meantime, we can work on the proper user-controlled migration functionality.

To make myself clear: I do agree that this functionality needs to be improved, but not at the cost of blocking things.

danielhollas:
@yakutovicha I don't think we had any migrations since the new Docker stack so I believe this warrants careful testing before merging.

Otherwise I agree there are no blockers.

unkcpz:
The CI test failed for `full-stack` because that is the one who tries to install `aiidalab-widgets-base` and `aiidalab-qe`. It failed when installing `aiidalab-qe` I guess there are some dependency conflicts. Let's wait for the new test, I'll check it after.

danielhollas:
@unkcpz one observation I made when looking at the test failure on GitHub (not directly related to this PR specifically): It seems that when the tests fail, the image is not published on ghcr.io as it used to before your refactoring. That makes debugging problematic since you can't download the built image to debug locally.

If my observation is correct, could you please move the image upload before the tests? Feel free to do it here or in a separate PR.

unkcpz:
Good point, I didn’t consider this situation. I hope add an always() can solve the problem. I’ll do it here since there is a failed test I can test for. 
Can you have a look at https://github.com/aiidalab/aiidalab-docker-stack/pull/412 if it is ready to be merge. Can save some build time to test things here.

danielhollas:
>  I hope add an always() can solve the problem.

Yeah that should work as well (you can check the syntax in AWB where I used this to always upload screenshots as artifacts). But why do you think this solution would be better than simply reordering the steps?

unkcpz:
> But why do you think this solution would be better than simply reordering the steps?

Because the action that push the image to ghcr is in next action job not the next step (which is uploading the artifact to github).

danielhollas:
Oh, okay. But what artifact is it uploading then? 😅

unkcpz:
Seems works after add `always()`, the image is https://github.com/aiidalab/aiidalab-docker-stack/pkgs/container/full-stack/139750905?tag=amd64-pr-396

> Oh, okay. But what artifact is it uploading then? 😅

It uploads the image saved (by `docker save`), and here are the artifacts https://github.com/aiidalab/aiidalab-docker-stack/actions/runs/6598466593?pr=396

danielhollas:
@unkcpz but uploading the artifact is the slow part no? That wasn't happening before, and I suspect if we simply push to ghcr.io straightaway it would be much more efficient. I don't think artifacts were designed for such large uploads, and I don't really see the benefits. 

Even if you want to push to both places, you can do it in a single job, instead of having to download the artifact and upload it to ghcr, that's very inefficient

danielhollas:
Also, normal users should pull from Docker, so the number of tags on ghcr.io is not a problem.

unkcpz:
Back to the failed test itself, the issue related to https://github.com/aiidalab/aiidalab/issues/397, I put here to cross link in case it is hard to find in the future. 
The "stable" aiidalab-qe in the test is `v23.04.6` which pining `aiida-core` with `aiida-core~=2.3.0` and will downgrade the `aiida-core` to `2.3.1` and cause the installation failed for the following test cases. 

In order to solve this problem I did two changes here:
1. towards https://github.com/aiidalab/aiidalab/issues/397, I pining the lower bounded `aiida-core` with the installed version to prevent it downgraded from apps. 
2. In the stable app version installation, I use `--pre` to use the released version without the tag-number version included.

unkcpz:
Just found another issue in terms of `aiidalab/aiidalab`, after https://github.com/aiidalab/aiidalab/commit/66710714bcee95602b210657d91a6e9455fa5541, we didn't make any release and we remove the `setup.py` for both qeapp and your app. It leads to when install the app by `aiidalab install` the warning "WARNING:aiidalab.app:Warning: App 'quantum-espresso' does not declare any dependencies." pop up. I'll make a release for `aiidalab/aiidalab`. Since `aiidalab` uses calendar versioning, I think it contains quite some changes and deserves to be `v23.10.0`, what do you think @yakutovicha @danielhollas ?

danielhollas:
> (EDIT: after a second thought, maybe make a aiidalab release of v23.03.2 could be more safe)

I don't think there have been many changes so this one seems more appropriate.

> Just found another issue in terms of aiidalab/aiidalab, after https://github.com/aiidalab/aiidalab/commit/66710714bcee95602b210657d91a6e9455fa5541, we didn't make any release and we remove the setup.py for both qeapp and your app.

Well, I did not if fact made this change, the PR for that is still open and marked as blocked. :-P https://github.com/ispg-group/aiidalab-ispg/pull/183
I am a little surprised that you did not notice that issue, because that means you did not install QeApp for quite some time...

unkcpz:
> I did not if fact made this change, the PR for that is still open and marked as blocked. 

Haha, I was too aggressive in the move. I found it was removed at https://github.com/aiidalab/aiidalab-qe/commit/397133437300c86874a58745c272625dbcad9656, although the commit is only for remove this single file but hide in the large refactoring PR https://github.com/aiidalab/aiidalab-qe/pull/439.

> I am a little surprised that you did not notice that issue, because that means you did not install QeApp for quite some time.

We install the QeApp quite often, probably twice a day in my case. The problem is for development, it was installed using pip directly. In the docker image preparation, the `aiidalab install` is used in the end but the dependencies installation was taken of by pip explicitly and didn't break the app. Plus "WARNING:aiidalab.app:Warning: App 'quantum-espresso' does not declare any dependencies." was raised as a warning so didn't get my notice. 

Two things I learned:
- A specific test is needed in qeapp to test install by `aiidalab install` and check the app is all working well.
- Specific purpose commit better to be separated as a PR (as @yakutovicha suggested before, maybe no need to have one commit per PR but for sure one purpose per PR).

unkcpz:
> Even if you want to push to both places, you can do it in a single job, instead of having to download the artifact and upload it to ghcr, that's very inefficient

I think we have been distracted a bit from various discussions here, I'll continue this discussion at https://github.com/aiidalab/aiidalab-docker-stack/pull/412 and hopefully, we can find the best solution and converge to a wiki post.

For bump the `aiida-core==v2.4.0` I think it is ready for review.

stack/base/Dockerfile
```diff
@@ -16,20 +16,21 @@ WORKDIR /opt/
...
      mamba-bash-completion \
-     --file requirements.txt \
+     --file /opt/requirements.txt \
      && mamba clean --all -f -y && \
      fix-permissions "${CONDA_DIR}" && \
      fix-permissions "/home/${NB_USER}"
 
-
-# Pin shared requirements in the base environemnt.
-RUN cat requirements.txt | xargs -I{} conda config --system --add pinned_packages {}
```
danielhollas, line 31: Why was this command removed?

stack/base/Dockerfile
```diff
@@ -16,20 +16,21 @@ WORKDIR /opt/
 
 ARG AIIDA_VERSION
 
+# Pin shared requirements in the base environemnt.
+# The lower bound of aiida-core is set to the version of aiida-core to prevent
+# lower aiida-core DB version installed and trigger database downgrade.
+RUN echo "pip==23.0.1" > /opt/requirements.txt && \
```
danielhollas, line 22: Let's also try to bump pip to the latest version while we are here. (I am honestly not sure if we need this constrain anymore, but I guess it is safer to have, but we should try to update it regularly. New pip version in particular should have a much faster dependency resolver.

Otherwise I agree with the minimum aiida-core constrain. For the future we should still think whether to use an exact constraint. It can still happen that one app upgrades aiida-core, and another one downgrades it and breaks the DB.

danielhollas (changes requested):
@unkcpz thanks! I agree, that discussion should happen elsewhere.

I know this kind of contradicts the policy, but I would also suggest to publish a new aiidalab package version and bump it here. I looked at the commit history and besides the setup.py there are no functional changes. I would then do careful testing of everything (instead of having to test multiple times). Feel free to disagree though.

stack/base/Dockerfile
unkcpz: Ah, by accident when I switched the lines. I added it back.

stack/base/Dockerfile
unkcpz: > bump pip to the latest version while we are here.

I remember there were some setuptools issues that blocked up updating it. I bump it and let's see if all works.

unkcpz: Change it to `23.3.1` and all tests passed.

unkcpz:
> I would then do careful testing of everything (instead of having to test multiple times). Feel free to disagree though.

If there is a test by hand, I don't know what reason I can find to disagree with it . Always better to have a human test.
I bumped the aiidalab to `23.3.2`, let's wait for the built and then test it. I'll do the test for the latest QEapp.

danielhollas:
I've tested install of my app, both with and without setup.py. I will test the DB migration tomorrow.

tests/test-full-stack.py
```diff
@@ -5,7 +5,7 @@
 def generate_aiidalab_install_output(aiidalab_exec, nb_user):
     def _generate_aiidalab_install_output(package_name):
         output = (
-            aiidalab_exec(f"aiidalab install --yes {package_name}", user=nb_user)
+            aiidalab_exec(f"aiidalab install --yes --pre {package_name}", user=nb_user)
```
danielhollas, line 8: Not sure if this is needed, since we're also installing directly from the branch in `test_install_apps_from_default_branch`. But I guess it is fine.

danielhollas (approved):
I've tested the DB migration and things seems to work. @unkcpz once you test with QeApp I think this is good to go. :rocket: 

Here's how I tested:
1. Launch a fresh container with `aiidalab/full-stack:latest` image, with aiida-core=2.3.1
2. Install my app and run some calculations
3. Stop the container
4. Edit profile configuration to use `aiidalab/full-stack:pr-396` 
5. Relaunch the container.
6. Check the container logs with `aiidalab-launch logs -p <test_profile>`
I saw this in the logs:
```
++ verdi storage migrate --force
Report: Migrating to the head of the main branch
Warning: Invalidating the hashes of certain nodes. Please run `verdi node rehash -e aiida.node:process.calculation.calcjob`.
Report: - main_0001 -> main_0002
Success: migration completed
++ verdi daemon start
Starting the daemon with 1 workers... OK
```
7. Verify in my app that the previous calculation is still there and that I can launch a new calculation without an issue.

tests/test-full-stack.py
unkcpz: Much appreciate for writing the detailed test flow, I'll test on QeApp with @superstar54 who has quite many containers with lots of calculations. 

Per `--pre` options I added, it is for QeApp specifically since we didn't make the release `v23.10.0` yet but the prerelease version. The stable version is still `v23.04.6` which pinning to `aiida-core~=2.3.0` will try to downgrade aiida-core and fail the test (exactly what we improved in this PR). 

I personally think here it is better to not use `--pre` but only test the stable version. I will keep this PR until we make the release of `v23.10.0` which the planned date is 6th November.

danielhollas: @unkcpz @superstar54 friendly ping, would be good to test and merge so that others can test this via the `edge` image.

unkcpz: Tested, all good. The migration is run and calculations can be reloaded with showing the same result properly.

unkcpz:
We found a small issue with updating the `pip` to `23.3.1`, where the fallback support using legacy `setup.py install` is removed, see https://github.com/pypa/pip/issues/8368. 
If an app version is not PEP 440 conformed, the installation was not failed but now it will fail.
The issue was reported by @AndresOrtegaGuerrero in MFA-CSCS app and fixed at https://github.com/aiidalab/aiidalab-mfa-cscs/pull/14

PR #397, opened 2023-07-10
Fix/container action triggering paths

unkcpz:
The workflow triggering paths are added


unkcpz:
After this one is merged, I'll make a release to test if the release actions are correct.

PR #408, opened 2023-09-07
Change to self-hosted runner in PSI for ARM64

unkcpz:
The self-hosted runner is deployed in PSI, it is a macOS M2 server that aims to test the docker build on ARM64 machine. The instructions on how this self-hosted runner is set up is create for reference.


unkcpz:
Hi @mbercx, I run all tests on my Macbook air, and seems all good. Can you give it a look and set up the self-host runner on your Mac when you get time? 

I use `colima` as the docker runtime instead of the standard Docker Desktop which I assume you will use for your daily work. This was adapted to avoid the images created and deleted by the CI actions having side effects on your daily work.

unkcpz:
This PR is blocked by https://github.com/aiidalab/aiidalab-qe/pull/449. Since I remove the permission of this repo to push to package for `aiidalab/qe` image. After https://github.com/aiidalab/aiidalab-qe/pull/449, I'll create PR in this repo to remove QE image build.

unkcpz:
After I tried this on my laptop, I ran into some issues with homebrew (since I reinstalled it), and it led me to this post https://www.codejam.info/2021/11/homebrew-multi-user.html.

I think to avoid the self-runner potentially breaking the regular system user (that is @mbercx in our case), I think the homebrew should installed only for Marnik, and the newly created `runner-user` should use sudo to run the script. But I may worried too much since the setup script only runs once to setup, and no need to be fully automatic.

PR #409, opened 2023-09-27
Vendor load-singlesshagent.sh script

danielhollas:
Couple changes here:

1. I think it is better to vendor the `load_singlesshagent.sh` script in this repo, rather than downloading it from a super-weird readthedocs URL.
2. Fixed an annoying spurious and confusing message from this script when you first enter the container and the `ssh-agent` is not yet running. (see third commit)
3. Enabled autocompletion for `aiidalab` CLI app.


unkcpz (commented):
@danielhollas Thanks! Fully agree to move the sshagent script here. Just to sure that you didn't change any code in this script correct? Other changes are all good to me.

danielhollas:
> Just to sure that you didn't change any code in this script correct? Other changes are all good to me.

I did actually, sorry for not making that clear. I removed the spurious error message (point 2. in the OP), see the third commit. Also in the last commit a made a small bugfix, substituting `$USER` to `$NB_USER`.

stack/base/load-singlesshagent.sh
```diff
@@ -32,7 +32,7 @@ load_singlesshagent() {
         [ "$VERBOSE" == "true" ] && echo "- no existing environment to source" >&2
     fi
 
-    SSH_ADD_OUTPUT=`ssh-add -l`
+    SSH_ADD_OUTPUT=`ssh-add -l 2> /dev/null`
```
danielhollas, line 35: @unkcpz here's the small change. Note that it is okay to redirect error to dev/null here since we check the return value of this `ssh-add` command below.

unkcpz:
> Also in the last commit a made a small bugfix, substituting $USER to $NB_USER.

Which means the original downloaded script should not work properly. I'll have a test to see how sshagent is actually working, I guess it is related to when ssh key phrase is not empty.

unkcpz:
@danielhollas thanks for the change. I read some detailed introductions on ssh-agent, along with the script. I think maybe the correct solution is to 
1. remove the `eval ssh-agen` since the script takes care of starting ssh-agent. 
2. Then move the `load-singlesshagent.sh` source section after the ssh key pairs are generated.  

Can you give this a try? Or if you don't mind I can add commits to test it.

danielhollas:
@unkcpz feel free to add commits with your proposed changes. Thanks!

danielhollas:
@unkcpz I am wondering: I think we still need the `eval ssh-agent`, since we want the ssh-agent to be started upon container startup, whereas I think `.bashrc` gets executed only when you enter the container? (maybe I am wrong on this). In which case, I am wondering if the singlesshagent script should be called directly, and not from `.bashrc`? I am not sure...

unkcpz:
Hi @danielhollas, thanks for bringing this up. I was testing exactly what you mentioned as well before lunch. 

> whereas I think .bashrc gets executed only when you enter the container? 

This is true. However, the `load_sshagent.sh` script also includes this and it already starts the `ssh-agent`. The script needs to be sourced so it makes sense to put it in `.bashrc`. 
What is missing previously is the `ssh-add` I think. You can try the new image now and if you go into the container you'll find `ssh-add -l` will show that the key is added to the list. 

```bash
docker pull ghcr.io/aiidalab/full-stack:sha-7a773b3
docker run --rm -it -p 8888:8888 ghcr.io/aiidalab/sha-7a773b3
```

Open the terminal (either from `docker exec` or from aiidalab terminal) and `ssh-add -l` to check the key is added.

I personally don't know why this `ssh-agent` is needed in our AiiDAlab case in the first place. We set the ssh private key with the empty passphrase, thus no need to input the passphrase even without `ssh-agent` to ssh to the remote once the passfree ssh is set up. 
From what I understand, the only benefit is the `ssh-agent` is helpful when there is a jump server, and then no need to add the private key to the jump server.
I guess @yakutovicha is more exporter on the SSH stuff, maybe he can see the proper way this was used and how it should be put in the container.

stack/base/load-singlesshagent.sh
```diff
@@ -0,0 +1,67 @@
...
+    if [ "$SSHADD_RETVAL" == "2" ]
+    then
+        [ "$VERBOSE" == "true" ] && echo "  - unable to contact agent, creating a new one" >&2
+        (umask 066; ssh-agent > ${SSH_ENV})
+        source "${SSH_ENV}" > /dev/null
+    elif [ "$SSHADD_RETVAL" == "1" ]
+    then
+        [ "$VERBOSE" == "true" ] && echo "  - ssh-agent found (${SSH_AGENT_PID}), no keys (I might want to add keys here)" >&2
+        # Maybe the user wants to do it by himself?
+        #ssh-add
```
danielhollas, line 48: @unkcpz I would prefer to uncomment this instead of adding it separately.

unkcpz: For us, I think yes. Is this you commented out or it was there already?

danielhollas:
I found this comment from @yakutovicha in https://github.com/aiidalab/aiidalab-docker-stack/pull/254#pullrequestreview-1078793632

> Another thing we should not forget is [the ssh-agent](https://github.com/aiidateam/aiida-prerequisites/blob/develop/bin/load-singlesshagent.sh) to be able to connect to machines that require ssh keys with password

For this use case, I don't know if automatically calling `ssh-add` works since the user needs to provide the passphrase.

We need more info from @yakutovicha to see how this is actually being used before we can continue here.

stack/base/load-singlesshagent.sh
unkcpz: > For this use case, I don't know if automatically calling ssh-add works since the user needs to provide the passphrase.

I see, then I think maybe calling it outside explicitly as I did in the last commit is a proper solution.

danielhollas:
@yakutovicha can you please clarify what is the current workload for private keys? Do we support private keys that require a passphrase? And if so how does the user provide the passphrase to get the key to ssh-agent?

unkcpz:
My two cents, see the reply for questions below.

> can you please clarify what is the current workload for private keys? 

The private key file can be uploaded by the `SshComputerWidget` and add a section in `.ssh/config` with the identityFile pointed to the path of the key. You found there are some issue on this when the key file has the same name of the private key existed, and it is fixed in https://github.com/aiidalab/aiidalab-widgets-base/pull/511

Otherwise, only the `id_rsa` which was generated the first time the container started is used.

> Do we support private keys that require a passphrase? 

No, I don't think so. But I think it would be a problem, user who has the private key with a passphrase means they generate it without using widget and has security concern about it so it is their duty to add the key to the ssh-agent and they can do it manually in terminal. 

Then it comes to the question, if we don't support passphrase, do we still need `ssh-agent`?
My answer is yes and in the AiiDAlab context we support only non-passphrase ssh key pairs.

The `ssh-agent` provides the following advantages, 1) if using `ssh-add` to put the private key to the agent, during the session the passphrase won't be asked anymore and with the passphrase the key is secure even it is stolen. **sadly** we didn't support passphrase in AiiDAlab. 2) ssh-agent can manage multiple keys, if you have multiple SSH keys for different purposes, `ssh-agent` can help you manage them without explicitly setting it in the `.ssh/config`. (we didn't use this feature either but it because necessary when we start to support MFA of CSCS, see the explain below).

The `aiidalab-mfa-cscs` app will generate `cscs-key` and `cscs-key.pub` and the key will expire in 1d. From the `ComputationalResourceWidget` it has no way to know the name of the private key that will be generated by other tools so it is only responsible for adding minimal information to the `.ssh/config` which has no `IdentityFile` option. With `ssh-agent`, `aiidalab-mfc-cscs` can do `ssh-add` to add the key it generated and then the connection is constructed. I have tested it locally and it works in charm. 

In summary, we need `ssh-agent` but we can not support private key with passphrase.

stack/base/load-singlesshagent.sh
unkcpz, line 48:
```suggestion
        ssh-add ~/.ssh/id_rsa
```
I'd argue to run this, we create `id_rsa`/`id_rsa.pub` pair in the container start script without passphrase, so we know it can be added without any problem. 
It also makes it possible to use ssh-agent for remote config that doesn't specify `IdentifyFile`.

stack/base/before-notebook.d/10_prepare-home-config.sh
```diff
@@ -40,12 +26,23 @@ fi
...
+# append the command text of source to .bashrc if the script /opt/bin/load-singlesshagen.sh is present
+# and the command text is not already present in .bashrc
+header="# Load singlesshagent on shell startup."
+if [[ -f /opt/bin/load-singlesshagent.sh ]] && ! grep -q "${header}" /home/${NB_USER}/.bashrc; then
+    cat >> "/home/${NB_USER}/.bashrc" <<- EOF
+
+${header}
+if [ -f /opt/bin/load-singlesshagent.sh ]; then
+    source /opt/bin/load-singlesshagent.sh
+    ssh-add /home/${NB_USER}/.ssh/id_rsa 2> /dev/null
```
unkcpz, line 45:
```suggestion
```
The key is added to agent by the load script above, so no need to do it explicitly here.

unkcpz:
@danielhollas I add two comments for what I think it is the proper way to deal with the agent and `ssh-add`. If you agree, we can apply the changes and then can test the image locally.

stack/base/before-notebook.d/10_prepare-home-config.sh
```diff
@@ -40,12 +26,23 @@ fi
...
 
-
 if [[ ! -f  /home/${NB_USER}/.ssh/id_rsa ]]; then
   # Generate ssh key that works with `paramiko`
   # See: https://aiida.readthedocs.io/projects/aiida-core/en/latest/get_started/computers.html#remote-computer-requirements
   ssh-keygen -f /home/${NB_USER}/.ssh/id_rsa -t rsa -b 4096 -m PEM -N ''
 fi
 
-# Start the ssh-agent.
-eval `ssh-agent`
```
danielhollas, line 51: Based on the discussion, I think we should source the `load-singlesshagent.sh` script here. Otherwise I don't think the ssh-agent will be running when the container starts, since .bashrc is not evaluated AFAIK.

stack/base/before-notebook.d/10_prepare-home-config.sh
danielhollas: Agreed.

stack/base/load-singlesshagent.sh
danielhollas: Yes, that sounds good. Please go ahead and change it and add a comment with the explanation.

danielhollas (commented):
@unkcpz thank you for looking into this and a detailed write up! See my comments, I agree with your plan about `ssh-add`

stack/base/before-notebook.d/10_prepare-home-config.sh
unkcpz: Ha! I think the one of the ssh-agent mentioned in https://github.com/aiidalab/aiidalab-mfa-cscs/pull/6 is from here! 
I agree to source the `load-singlesshagent.sh` here. It avoids the duplicate `ssh-agent` I think.

danielhollas: Yeah, interesting. That would mean that it is started somewhere else before we get to this script. Maybe the Jupyter image that we use already starts it somewhere. In any case, source the script here is the right thing to do I think.

unkcpz:
@danielhollas should all work as expected now. I also added a test to check the ssh-agent is started properly. 
Please have a test on the image after it is successfully built.

stack/base/load-singlesshagent.sh
```diff
@@ -0,0 +1,67 @@
...
+    SSH_ENV="$HOME/.ssh/agent-environment"
+    # Source SSH settings, if applicable
+    if [ -r "${SSH_ENV}" ]; then
+        source "${SSH_ENV}" 2> /dev/null
+        [ "$VERBOSE" == "true" ] && echo "- sourcing existing environment" >&2
+    else
+        [ "$VERBOSE" == "true" ] && echo "- no existing environment to source" >&2
+    fi
+
+    SSH_ADD_OUTPUT=`ssh-add -l`
```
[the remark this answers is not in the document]
danielhollas: Huh, that is really strange, are you sure? I am only redirecting stderr, that should AFAIK not change the exit code.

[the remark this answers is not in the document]
unkcpz: I am not sure 😛 if I don’t remove it the ssh-agent create clause is not hit. I’ll play more to figure out why. But overall everything works fine no matter if the output is redirected or not.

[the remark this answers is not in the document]
unkcpz: I changed it back to `ssh-add -l 2> /dev/null` and it still works fine. The problem I encountered is not because of this line but the other redirected probably. 

Then this PR is ready, please have another look. @danielhollas

PR #416, opened 2023-11-14
Use aiidalab wide logo in notebook interface

unkcpz:


stack/lab/Dockerfile
```diff
@@ -105,3 +105,9 @@ ENV NOTEBOOK_ARGS \
      "--MappingKernelManager.cull_interval=300" \
      "--TerminalManager.cull_inactive_timeout=600" \
      "--TerminalManager.cull_interval=60"
+
+# Set up the logo of notebook interface
+COPY --chown=${NB_UID}:${NB_GID} aiidalab-wide-logo.png ${CONDA_DIR}/notebook-logo.png
```
danielhollas, line 110: Why do you put the logo in `$CONDA_DIR`? Isn't `/opt/` a better place? Perhaps `/opt/static/`?

stack/lab/Dockerfile
```diff
@@ -105,3 +105,9 @@ ENV NOTEBOOK_ARGS \
...
+# the folder of logo.png is get by find "$CONDA_DIR/lib" -path "*/notebook/static/base/images/logo.png"
```
danielhollas, line 111: I am very confused by this code. Why is this `find` necessary? Isn't the original logo always in the same path?

danielhollas (changes requested):
Could you please add a screenshot? Where is this logo going to show up in the UI?

stack/lab/Dockerfile
unkcpz: Could be, I'll use `/opt`, it is a temporary location, and I think it is better to use `mv` instead of `cp`.

stack/lab/Dockerfile
unkcpz: > Isn't the original logo always in the same path?

Not exactly the path now is `/opt/conda/lib/python3.9/site-packages/notebook/static/base/images/logo.png`, thus also dependent on the python version.

unkcpz (commented):
The logo appear on the left-up corner replace the "Jupyter" logo, now it look like.

image

stack/lab/Dockerfile
danielhollas: Right, but we control the python version and it's value is available in the Dockerfile you only need to somehow get rid of the patch version.e.g 3.9.13 to 3.9, but I am assuming that should be possible.

stack/lab/Dockerfile
unkcpz: Can't use `/opt` since I didn't run `fix-permission`. I use `/tmp` instead, so I can use `mv` to avoid

stack/lab/Dockerfile
unkcpz: Did not manage to do it simply. 

I can do:
```
RUN export PYTHON_VERSION_MAJOR_MINOR=$(python -c "import sys; print('python{}.{}'.format(*sys.version_info[:2]))") && \
    mv /tmp/notebook-logo.png "$CONDA_DIR/lib/$PYTHON_VERSION_MAJOR_MINOR/site-packages/notebook/static/base/images/logo.png"
```

But I don't think it is simple. If read from `PYTHON_VERSION` then the problem is it can be `3.9` or `3.9.13`, so it is not easy to parse the `<major>.<minor>` from it.

unkcpz: I add a comment to explain the usage of `find`, I assume the `-exec` is a common option to execute the command on the result of `find` command.

stack/lab/Dockerfile
```diff
@@ -105,3 +105,11 @@ ENV NOTEBOOK_ARGS \
...
      "--TerminalManager.cull_inactive_timeout=600" \
      "--TerminalManager.cull_interval=60"
+
+# Set up the logo of notebook interface
+COPY --chown=${NB_UID}:${NB_GID} aiidalab-wide-logo.png /tmp/notebook-logo.png
+
+# The directory location of logo.png is in the `/opt/conda/lib/python3.9/site-packages/notebook/static/base/images/logo.png`, but the python version may change
+# I use the find to find the location and mv the logo.png to the folder
+# copy the logo.png to the folder as logo.png
+RUN find "$CONDA_DIR/lib" -path "*/notebook/static/base/images/logo.png" -exec mv /tmp/notebook-logo.png {} \;
```
danielhollas, line 115: I still think the current solution is overcomplicated. You should be able to use wildcard expansion, something like this?

```suggestion
RUN mv /tmp/notebook-logo.png /opt/conda/lib/python3*/site-packages/notebook/static/base/images/logo.png
```

Also, for the `PYTHON_VERSION`, we will always specify the full version, `major.minor.patch`, because the patch version also specifies the version of Jupyter dependencies.

danielhollas (changes requested):

stack/lab/Dockerfile
unkcpz: Cool, I didn't know this will work. I test locally and seems work! Let me adapt your suggestion and test on image.

danielhollas: Note, in BASH this parameter expansion could be used to extract 3.9 from 3.9.12

```bash
${PYTHON_VERSION/%.+([0-9])}
```

This assumes that `extglob` option is enabled, which at least on Ubuntu seems to be the case. Otherwise can be enabled with 

```bash
shopt -s extglob
```

https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html
https://www.gnu.org/software/bash/manual/html_node/Pattern-Matching.html

So this might work as well

```bash
RUN shopt -s extglob && mv /tmp/notebook-logo.png /opt/conda/lib/python${PYTHON_VERSION/%.+([0-9])}/site-packages/notebook/static/base/images/logo.png
```

unkcpz: > because the patch version also specifies the version of Jupyter dependencies.

This is the decision we made in `aiidalab-docker-stack`, jupyter docker stacks support both with/without patch version.

I think using the wildcard is simple and works for this case, I'll just use it.

PR #418, opened 2023-12-12
Relax pyyaml dev dependency

danielhollas:
+ remove `requirements-dev-arm64.txt` which was just a duplicate of `requirements-dev.txt`

This should shut up Dependabot reporting vulnerability in the old pyyaml version. We needed to restrict `pyyaml` due to some broken dependency, hopefully that has been resolved by now.


unkcpz (approved):
Thanks! That's true, it is duplicated (was not, because I need it to try to pass the integrating tests).

PR #419, opened 2023-12-15
Bump actions/artifact-upload to v4

danielhollas:
This should significantly speed up artifact upload, see https://github.com/actions/upload-artifact#v4---whats-new

TODO: Probably we should turn off compression, see https://github.com/actions/upload-artifact#altering-compressions-level-speed-v-size

### DIFF
--- a/.github/actions/create-dev-env/action.yml
+++ b/.github/actions/create-dev-env/action.yml
@@ -18,7 +18,7 @@ runs:
           if: ${{ inputs.architecture == 'amd64' }}
           uses: actions/setup-python@v4
           with:
-              python-version: 3.x
+              python-version: 3.11
 
         - name: Install Dev Dependencies 📦
           run: |
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -213,7 +213,7 @@ jobs:
         runs-on: ubuntu-latest
         needs: [merge-tags-ghcr, merge-tags-dockerhub]
         steps:
-            - uses: actions/checkout@v3
+            - uses: actions/checkout@v4
 
             - name: Create release
               uses: softprops/action-gh-release@v1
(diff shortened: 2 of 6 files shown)


.github/workflows/docker.yml
```diff
@@ -3,53 +3,13 @@ name: Build, test and push Docker Images
 
 on:
     pull_request:
-        paths:
-            - ".github/workflows/docker.yml"
```
danielhollas, line 7: This seemed to include essentially all files and seemed very fragile. For now I simply excluded Markdown files, although I tempted to even remove that.

.github/actions/create-dev-env/action.yml
```diff
@@ -18,7 +18,7 @@ runs:
           if: ${{ inputs.architecture == 'amd64' }}
           uses: actions/setup-python@v4
           with:
-              python-version: 3.x
+              python-version: 3.11
```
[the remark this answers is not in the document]
unkcpz: Yep, agree.

unkcpz:
Thanks! I just noticed the update. It makes upload much faster! The test failed because of Marnik's machine was shutdown, now should build and test on arm64.

> Probably we should turn off compression

How much time it will save without compress?

.github/workflows/docker.yml
unkcpz: Makes a lot sense for this repo, go ahead.

danielhollas:
> How much time it will save without compress?

Seems to save a lot. Even the lowest level of compression (lvl 1) is ~20s slower. Note the download-artifact still takes around 1 minute, and there are other extra steps (saving and loading Docker images) that make all this slower.

My apologies, I have by accident pushed some commits on master, so I needed to merge this without approval. Please leave a review if you have any comments and I'll make a new PR.

PR #421, opened 2023-12-16
Downgrade artifact-upload to fix arm64 build?

danielhollas:
The arm64 build is failing on artifact-upload step. I am not sure if that is the consequence of the bump of `actions/artifact-upload` to v4, or something is wrong with our self-hosted runners. Let's see if downgrading helps.

### DIFF
--- a/.github/actions/create-dev-env/action.yml
+++ b/.github/actions/create-dev-env/action.yml
@@ -18,7 +18,7 @@ runs:
           if: ${{ inputs.architecture == 'amd64' }}
           uses: actions/setup-python@v4
           with:
-              python-version: 3.x
+              python-version: 3.11
 
         - name: Install Dev Dependencies 📦
           run: |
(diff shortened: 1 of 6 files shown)


danielhollas:
Looks like we will have to downgrade the `actions/upload-artifact` back to v3 since the v4 does not work on our self-hosted runner yet.

In the [release notes](https://github.com/actions/upload-artifact#breaking-changes), they mention this:
> On self hosted runners, additional [firewall rules](https://github.com/actions/toolkit/tree/main/packages/artifact#breaking-changes) may be required.

cc @mbercx @unkcpz 

Before this is debugged, I'd suggest to merge this to get us working again.

unkcpz:
Thanks for taking care of `artifact-upload` action. If it break the github workflow, we have to postpone the update of it. 
And I'll work on not uploading the artifact this week.

unkcpz (approved):

unkcpz:
One more thing that needs to be noticed is running on @mbercx's machine may not the final solution, since he found somehow it maybe the reason that caused his daily AiiDA work environment break.

danielhollas:
Closed in favour of #423

PR #423, opened 2023-12-18
Bump actions/artifact-upload to v4

unkcpz:
This is a revert amendment of https://github.com/aiidalab/aiidalab-docker-stack/pull/419 which merged improperly with test commints.

- The changes should significantly speed up artifact upload, see https://github.com/actions/upload-artifact#v4---whats-new
- Also include the updates of `action/checkout` action to v4
- include fix the python version of running environment to `3.11`.
- include the changes of removing paths change detection in `docker.yml`

### DIFF
--- a/.github/actions/create-dev-env/action.yml
+++ b/.github/actions/create-dev-env/action.yml
@@ -18,7 +18,7 @@ runs:
           if: ${{ inputs.architecture == 'amd64' }}
           uses: actions/setup-python@v4
           with:
-              python-version: 3.x
+              python-version: 3.11
 
         - name: Install Dev Dependencies 📦
           run: |
--- a/.github/actions/load-image/action.yml
+++ b/.github/actions/load-image/action.yml
@@ -16,7 +16,7 @@ runs:
     using: composite
     steps:
         - name: Download built image 📥
-          uses: actions/download-artifact@v3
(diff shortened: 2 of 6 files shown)


unkcpz:
I am ordering a request for the firewall rule in PSI, hope it will work. I'll keep this PR open for the test. 
Will also look into the timeout of the action.

danielhollas:
Thanks @unkcpz. If the firewall rule does not fix it, we can use V3 action for arm and v4 for amd, and then focus on removing the artifacts entirely.

unkcpz:
> If the firewall rule does not fix it, we can use V3 action for arm and v4 for amd, and then focus on removing the artifacts entirely.

Sadly no, from IT he said the TCP ports 80/443 is allowed by default, therefore no specific firewall rule is needed.

unkcpz:
I test it with aiida-core image which is a smaller one, all good. I think it is the problem of the action it self. See https://github.com/aiidateam/aiida-core/pull/6228

unkcpz:
I remove the compression setting and all works. There are two possibilities, 1) the artifact is too large 2) no compression causes the problem.

unkcpz:
> then focus on removing the artifacts entirely.

I'll find time work on this this week anyway.

danielhollas (approved):

PR #427, opened 2024-01-17
Bump aiida-core to 2.4.3 to include the fix of mpi setup

unkcpz:


unkcpz:
Need to wait for aiida-core docker stack is available. This is one small reason I don't like using conda to install the core dependencies, with pypi we can always get things immediately.

danielhollas:
> Need to wait for aiida-core docker stack is available. This is one small reason I don't like using conda to install the core dependencies, with pypi we can always get things immediately.

True, but the couple days wait is not worth it IMHO. Also, installing via pip would most likely increase the image size since conda is more efficient for bundling non-Python dependencies.

unkcpz:
> True, but the couple days wait is not worth it IMHO. 

Okay, not hard to have conda feedstock update manually. Let's keep on using conda, we live quite well with it.

danielhollas (approved):
@yakutovicha can you please test that this fixes your MPI issues?
(image `aiidalab/full-stack:pr-427`)

unkcpz:
Discussed with @yakutovicha on slack, we agree on get this merged and ask @cpignedoli test on the new image.

Issue #429, opened 2024-02-04
Migrate to GHA M1 runners


danielhollas:
Currently we use our own runners for the Arm build. GitHub recently made M1 runners available. Switching to them should hopefully significantly streamline our build.

I think this should be done first, and will hopefully make it easier to then remove the `artifact-upload` steps, as discussed in #414 

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

See `macos-14` runner.

unkcpz:
Wow! This is fantastic, I checked 2 months ago and it cost quite a lot. Thanks for the update.

danielhollas:
I don't think this is relevant to us but just in case:
https://x.com/charliermarsh/status/1754542590517551258?s=20

yakutovicha:
Late to the party, but happy to see GitHub finally enabled the ARM runners. That can simplify the image build pipeline

unkcpz:
Sadly, the runner is has some virtualization setting not support running docker inside yet.

danielhollas:
I am wondering, how does it work on the self-hosted runner? Is the runner running outside of Docker so it does not require nested virtualization?

unkcpz:
From what I understand, the different between the self-hosted runner and the github runner is the github runner is already a virtulized machine. It not supported to have another virtualization layer inside it.
While for self-hosted machine, the machine itself is a physical machine and running docker through colima is just one layer virtulization.

danielhollas:
Looks like there isn't much we can do here for foreseeable future so I'll close this.
I subscribed to this issue in case something changes in the future https://github.com/douglascamata/setup-docker-macos-action/issues/26
But it seems this would need changes from both Github (using >=M2 runners) and Apple (providing software support for nested virtualization). Too bad. :-(

PR #430, opened 2024-02-23
Try: move to macos-14 in GHA

unkcpz:
address #429 

Let's see if it works.


danielhollas:
Looks like there isn't much we can do here for foreseeable future so I'll close this. 
I subscribed to this issue in case something changes in the future https://github.com/douglascamata/setup-docker-macos-action/issues/26
But it seems this would need changes from both Github (using >=M2 runners) and Apple (providing software support for nested virtualization). Too bad. :-(

PR #431, opened 2024-02-26
Add Dependabot config for updating GHAs

danielhollas:


danielhollas:
I've added a pin for pytest to fix the build on the self-hosted runner, but this really should be fixed properly in #432.

@unkcpz will you have time to work on that in a reasonable future?

Issue #432, opened 2024-02-27
Install dev dependencies in separate environment on self-hosted runner


danielhollas:
I saw a very strange test failure in #431 

https://github.com/aiidalab/aiidalab-docker-stack/actions/runs/8054045469/job/22002214446?pr=431

Pytest fails in a very peculiar way inside inside `pytest-selenium` plugin. But we do not use this plugin in the repo!!!

Looking at the `.github/actions/create-dev-env/action.yml`, it seems that we're installing via pip into some global folder? This seems very bad. We should probably create a venv and install things locally. Something like `python -m venv && source .venv/bin/activate && pip install -r requirements-dev.txt`.

unkcpz:
The environment is set when the runner user started, so every time the action is run in with the environment activate.
I do it by having `runsvc.sh` (this is the script start the self-host runner) as:

```
➜  aiidalab-actions-runner cat runsvc.sh 
#!/bin/bash

# convert SIGTERM signal to SIGINT
# for more info on how to propagate SIGTERM to a child process see: http://veithen.github.io/2014/11/16/sigterm-propagation.html
trap 'kill -INT $PID' TERM INT

if [ -f ".path" ]; then
    # configure
    export PATH=`cat .path`
    echo ".path=${PATH}"
    source .venv/aiidalab-runner/bin/activate
    export PATH=$HOME/.venv/aiidalab-runner/bin:$PATH
    export DOCKER_HOST="unix://$HOME/.colima/aiidalab/docker.sock"
fi

nodever=${GITHUB_ACTIONS_RUNNER_FORCED_NODE_VERSION:-node16}

# insert anything to setup env when running as a service
# run the host process which keep the listener alive
./externals/$nodever/bin/node ./bin/RunnerService.js &
PID=$!
wait $PID
trap - TERM INT
wait $PID
```

But I think to make it more clear, this source activate better to move in the action. Because we use this environment for both the `aiidalab-docker-stack` and `aiidalab-qe`. The python environment although it is the same for the moment, but should be independent.

danielhollas:
> Because we use this environment for both the aiidalab-docker-stack and aiidalab-qe. The python environments are the same for both at the moment, but should be independent.

Oh, that's definitely not good. And I think even in the same repo, each individual jobs should actually have their own environment, otherwise it's very easy for different PRs to step on each other. 

[...]

PR #437, opened 2024-03-19
Set PIP_USER in base image

### LINKED ISSUE #435: Set PIP_USER=true in the image
When using pip to install packages, we always user the `--user` flag so that the packages are installed in `~/.local/` and thus saved in a docker volume that backs the home directory.

We should just make it the default, so that in case user runs pip directly for whatever reason, they automatically install to `~/.local/` even if they forget to specify `--user`. (This happened to me personally **many** times)

See:
https://pip.pypa.io/en/stable/topics/configuration/#environment-variables

danielhollas:
Ensure that pip installs to `~/.local` by default, instead of `/opt/conda`. 

Couple other changes:
 - Do not pin pip version and always upgrade to latest, as recommended by pip maintainers. If there are any breaking changes in the future, they should be caught by the integration tests.
 - Install appmode from PyPI, we install it from Github only for historical reasons AFAIK.
 - Make `--variant` parameter of pytest required
 - Minor cleanup in tests
Closes #435

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -83,6 +83,9 @@ target "base-with-services" {
     "PGSQL_VERSION" = "${PGSQL_VERSION}"
   }
 }
+# PYTHON_MINOR_VERSION is a Python version string
+# without the patch version (e.g. "3.9")
+# Used to construct paths to Python site-packages folder.
 target "lab" {
   inherits = ["lab-meta"]
   context = "stack/lab"
@@ -93,6 +96,7 @@ target "lab" {
   args = {
     "AIIDALAB_VERSION"      = "${AIIDALAB_VERSION}"
     "AIIDALAB_HOME_VERSION" = "${AIIDALAB_HOME_VERSION}"
+    "PYTHON_MINOR_VERSION" = join(".", slice(split(".", "${PYTHON_VERSION}"), 0, 2))
   }
 }
 target "full-stack" {
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -16,11 +16,10 @@ WORKDIR /opt/
 
 ARG AIIDA_VERSION
 
-# Pin shared requirements in the base environemnt.
+# Pin shared requirements in the base environment.
 # The lower bound of aiida-core is set to the version of aiida-core to prevent
 # lower aiida-core DB version installed and trigger database downgrade.
-RUN echo "pip==23.3.1" > /opt/requirements.txt && \
-     echo "aiida-core>=${AIIDA_VERSION},<3" >> /opt/requirements.txt
+RUN echo "aiida-core>=${AIIDA_VERSION},<3" > /opt/requirements.txt
 
 # Install the shared requirements.
 RUN mamba install --yes \
@@ -35,7 +34,9 @@ RUN mamba install --yes \
 RUN cat /opt/requirements.txt | xargs -I{} conda config --system --add pinned_packages {}
 
 # Configure pip to use requirements file as constraints file.
-ENV PIP_CONSTRAINT=/opt/requirements.txt
+ENV PIP_CONSTRAINT /opt/requirements.txt
+# Ensure that pip installs packages to ~/.local by default
+ENV PIP_USER 1
 
 # Enable verdi autocompletion.
 RUN mkdir -p "${CONDA_DIR}/etc/conda/activate.d" && \
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -30,9 +30,11 @@ RUN mamba install --yes \
      fix-permissions "${CONDA_DIR}" && \
      fix-permissions "/home/${NB_USER}"
 
-# Pin shared requirements in the base environemnt.
+# Pin shared requirements in the conda base environment.
 RUN cat /opt/requirements.txt | xargs -I{} conda config --system --add pinned_packages {}
 
+# Upgrade pip to latest
+RUN pip install --upgrade --no-cache-dir pip
 # Configure pip to use requirements file as constraints file.
 ENV PIP_CONSTRAINT /opt/requirements.txt
 # Ensure that pip installs packages to ~/.local by default
--- a/stack/lab/Dockerfile
+++ b/stack/lab/Dockerfile
@@ -52,7 +52,7 @@ RUN pip install appmode==0.8.0 --no-cache-dir --no-user
 RUN jupyter nbextension enable --py --sys-prefix appmode && \
      jupyter serverextension enable --py --sys-prefix appmode
 
-# Use an AiiDAlab gears icon instead of appmode one during
+# Swap appmode icon for AiiDAlab gears icon, shown during app load
 COPY --chown=${NB_UID}:${NB_GID} gears.svg ${CONDA_DIR}/share/jupyter/nbextensions/appmode/gears.svg
 
 # Copy start-up scripts for AiiDAlab.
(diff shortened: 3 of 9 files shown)


stack/base/Dockerfile
```diff
@@ -16,11 +16,10 @@ WORKDIR /opt/
 
 ARG AIIDA_VERSION
 
-# Pin shared requirements in the base environemnt.
+# Pin shared requirements in the base environment.
 # The lower bound of aiida-core is set to the version of aiida-core to prevent
 # lower aiida-core DB version installed and trigger database downgrade.
-RUN echo "pip==23.3.1" > /opt/requirements.txt && \
-     echo "aiida-core>=${AIIDA_VERSION},<3" >> /opt/requirements.txt
+RUN echo "aiida-core>=${AIIDA_VERSION},<3" > /opt/requirements.txt
```
danielhollas, line 22: I'll pin aiida-core version in a separate PR.

danielhollas:
@unkcpz @mbercx   it looks like docker daemon is not running on the ARM64 self-hosted runner.

https://github.com/aiidalab/aiidalab-docker-stack/actions/runs/8348422726/job/22850363649?pr=437

danielhollas:
Arm runners are running again!

mbercx:
> Arm runners are running again!

Yeah, sorry about that! Not sure why my work station keeps on shutting down from time to time. 🤷

unkcpz:
> Arm runners are running again!

Somehow there was an issue on the colima to start the aiidalab docker instance, and I delete the profile and restart. I think it bring the issue https://github.com/aiidalab/aiidalab-docker-stack/issues/432 to high priority.

unkcpz:
> Install appmode from PyPI, we install it from Github only for historical reasons AFAIK.

That's true.

stack/lab/Dockerfile
```diff
@@ -41,19 +41,17 @@ ARG AIIDALAB_HOME_VERSION
 RUN git clone https://github.com/aiidalab/aiidalab-home && \
      cd aiidalab-home && \
      git checkout v"${AIIDALAB_HOME_VERSION}" && \
-     pip install --quiet --no-cache-dir "./" && \
+     pip install --no-user --quiet --no-cache-dir "./" && \
```
unkcpz, line 44: With `--no-user` the package will be installed to where?

danielhollas: `--no-user` simply means default pip behaviour, so in this case it will be installed to `/opt/conda`, as before.

stack/base/Dockerfile
```diff
@@ -31,11 +30,15 @@ RUN mamba install --yes \
...
+# Pin shared requirements in the conda base environment.
 RUN cat /opt/requirements.txt | xargs -I{} conda config --system --add pinned_packages {}
 
+# Upgrade pip to latest
+RUN pip install --upgrade --no-cache-dir pip
 # Configure pip to use requirements file as constraints file.
-ENV PIP_CONSTRAINT=/opt/requirements.txt
+ENV PIP_CONSTRAINT /opt/requirements.txt
+# Ensure that pip installs packages to ~/.local by default
+ENV PIP_USER 1
```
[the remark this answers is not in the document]
danielhollas: Yeah, sorry, this is actually equivalent to providing `--user` on the CLI.
You can set any pip argument via envvar by capitalizing it and pre-fixing with `PIP_`

stack/lab/Dockerfile
danielhollas: https://pip.pypa.io/en/stable/topics/configuration/#environment-variables

docker-bake.hcl
```diff
@@ -93,6 +96,7 @@ target "lab" {
   args = {
     "AIIDALAB_VERSION"      = "${AIIDALAB_VERSION}"
     "AIIDALAB_HOME_VERSION" = "${AIIDALAB_HOME_VERSION}"
+    "PYTHON_MINOR_VERSION" = join(".", slice(split(".", "${PYTHON_VERSION}"), 0, 2))
```
unkcpz, line 99: Thanks for also take care of this.

stack/lab/Dockerfile
```diff
@@ -41,20 +41,20 @@ ARG AIIDALAB_HOME_VERSION
...
-COPY gears.svg ./appmode/appmode/static/gears.svg
-RUN pip install ./appmode --no-cache-dir && \
-     jupyter nbextension enable --py --sys-prefix appmode && \
+# Install and enable appmode, turning Jupyter notebooks to Apps
+RUN pip install appmode==0.8.0 --no-cache-dir --no-user
+# Enable appmode extension
+RUN jupyter nbextension enable --py --sys-prefix appmode && \
      jupyter serverextension enable --py --sys-prefix appmode
 
+# Use an AiiDAlab gears icon instead of appmode one during
```
unkcpz, line 55:
```suggestion
# Use an AiiDAlab gears icon instead of appmode one during loading app.
```

I guess you mean this?

tests/conftest.py
```diff
@@ -54,14 +64,15 @@ def execute(command, user=None, **kwargs):
             command = f"exec -T --user={user} aiidalab {command}"
         else:
             command = f"exec -T aiidalab {command}"
-        return docker_compose.execute(command, **kwargs)
+        out = docker_compose.execute(command, **kwargs)
+        return out.decode()
```
unkcpz, line 68: Great, this makes things more clear.

danielhollas:
@unkcpz thank for taking a look. I've pushed two new tests for the new pip behaviour. If you agree, I'll merge this and release a new latest version.

danielhollas:
@unkcpz I've merged this, feel free to take a look at the new tests as part of reviewing #442

tests/conftest.py
```diff
@@ -54,14 +64,30 @@ def execute(command, user=None, **kwargs):
...
+    package = None
+
+    def _pip_install(pkg, **args):
+        nonlocal package
+        package = pkg
+        return aiidalab_exec(f"pip install {pkg}", **args)
+
+    yield _pip_install
+    if package:
+        aiidalab_exec(f"pip uninstall --yes {package}")
```
unkcpz, line 90: I remember I mentioned somewhere there is an issue that the uninstallation will only remove the package but not the dependencies. Is this fixture solve that?

unkcpz (commented):
All good, thanks for working on this @danielhollas 
I had one question on the `pip_install` fixture.

tests/conftest.py
danielhollas: Good point, it doesn't. I am not sure if this is a problem in practice: Here I only use it for a package that I specifically chosen to have zero dependencies. But it would be good to at least document this.

I think it was a slightly bigger issue when running the integration tests and installing actuall AiiDAlab apps.

unkcpz: > I think it was a slightly bigger issue when running the integration tests and installing actuall AiiDAlab apps.

You mean we should not test the QE-app or AWB in the full-stack image? That what I proposed but you said you were to comfortable with a image that not working with QeApp. 
Can we make it a nightly test for QeApp and AWB that runs periodically?

unkcpz: > But it would be good to at least document this.

Yep, I'll add a comment on it.

danielhollas: >  Can we make it a nightly test for QeApp and AWB that runs periodically?

I don't think that is very developer friendly, I'd much rather now about issues rightaway. 

I will note however that originally these app-installation tests were completely separate from the others and marked as "integration" tests, they specifically were not supposed to block merging. We should go back to that, I'll do that as part of #439, or as a separate PR after #442 is merged, which is a pre-requisite for this work.

PR #439, opened 2024-03-20
Refactor CI workflow

### LINKED ISSUE #414: Slow build test and push
The build, test, and push to ghcr.io for each PR is still slow, the reason is the image is pushed to github artifacts. 
It can be accelerated by pushing to ghcr.io without uploading the artifacts.

danielhollas:
## Overview 

The goal of the PR is twofold:

- Improve CI runtimes (currently the overall workflow takes ~1h40min!)
- Reduce complexity

In the end, I mostly ended up reverting things back to where they were roughly around PR #372, while (hopefully) keeping the current modularity. Closes #414 

## Major changes

- The main change here is to stop uploading Docker images as artifacts: Instead we upload them to ghcr.io straightaway. This already brings a lot of simplification and time reduction -- the only "disadvantage" is that the PRs cannot be open from forks, but as discussed in #414, this is more of a feature rather than a bug.

- The second approach I took was to reduce our reliance on the self-hosted ARM64 runner, which is a limited resource (can only run one job at a time). In this PR, the arm64 image is build on ubuntu GHA runner with the help of QEMU virtualization. This is a bit slower then building natively, but not by that much (12min versus 9min). The other big advantage of this is that it reduces a lot of complexity, since we don't need to create / manage manifests manually, and simply let the `docker/bake-action` take care of the multiplatform build and upload. We can re-evaluate this in a future PR.

- We first build the amd64 images and immediately test `full-stack`. This overall takes only 6 minutes, so any obvious issues should be caught early.
I believe some of the improvements here should be ported to QEApp and aiida-core repos as well.

## Final results
- The workflow now takes 20 minutes. The first amd64 build and full-stack tests take only around 6 minutes, so any issues are surfaced early.
- The overall diff here in terms of lines (+551/-502) is a bit misleading, since I've added a lot of explanation and example inputs / outputs in the auxiliary shell scripts. The total lines of the yaml files is actually ~~+352/-470~~ +320/-470. 

## Future work

- Revamp release procedure so that we do not have to push to `main` 
- Reconsider buidling of arm64 image on self-hosted runner, if GHA turns out to be unreliable. On some commits on this PR the build took over 1h instead of 12 minutes.
- Speed up the build by optimizing what we put in the image. I have some ideas but I'll keep them for a separate PR.


unkcpz:
Thanks! 
Just careful that I don't want to go back to the original situation that tests for all 4*2 images are just in one GH workflow (or two for two architectures). The bake can happened in one workflow but tests are better separated.

danielhollas:
Yes, tests should be run separately and in parallel. But the current build strategy is extremely wasteful, since we're not reusing the previous layers (e.g. the `base` image is build 4 times, unless I am missing something).

unkcpz:
> (e.g. the base image is build 4 times, unless I am missing something).

That is correct! Initially I want to keep things as separated as possible and notice in the end every build cost < mins so didn't put effort to combine the built processes back to one. 
It definitely worth to try. Let me know if you encounter any problem and sorry I delayed on starting to work on this.

danielhollas:
@mbercx would you mind installing `jq` on the ARM runner? Thx

.github/workflows/build.yml
```diff
@@ -0,0 +1,148 @@
...
+        steps:
+
+            - name: Checkout Repo ⚡️
+              uses: actions/checkout@v4
+
+            - uses: docker/login-action@v2
+              with:
+                  registry: ghcr.io
+                  username: ${{ github.actor }}
+                  password: ${{ secrets.GITHUB_TOKEN
```
danielhollas, line 123:
```suggestion
                  password: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,137 @@
...
+                  cache: pip
+                  cache-dependency-path: |
+                      **/requirements*.txt
+
+            - name: Install dependencies
+              run: pip install -r requirements-dev.txt
+
+            - name: Run tests
+              # TODO: Hack with REGISTRY
+              run: REGISTRY= pytest -s --variant ${{matrix.target}} tests/test-common.py
```
danielhollas, line 136:
```suggestion
              run: REGISTRY= pytest -sv --variant ${{matrix.target}} tests/test-common.py tests/test-${{matrix.target}}.py
```

unkcpz:
On the mac machine I think it is better to be installed by homebrew, then only @mbercx has access to do it.

danielhollas:
@unkcpz It would be great if you could take a look now to see if the approach here makes sense to you, now that the amd64 build and tests are working.

What I did so far:
- Remove artifact upload / download
- Build all images in a single step (for amd64, building all 4 images takes only 3 minutes).
- After successful build the images are pushed to ghcr.io, tagged with a commit sha and PR number
- Run tests as separate jobs. amd64 tests take about 2 minutes each (but are run in parallel)
- Separated integration tests (qeapp and AWB installations) into a separate job
- Given that overall amd64 build and tests take only around 5 minutes, I think it makes sense to only trigger ARM64 build/test after all tests pass to save resources on our self-hosted runner

What remains to be done:
- ARM64 build and tests (waiting for jq)
- Pushing to DockerHub

.github/workflows/build.yml
```diff
@@ -0,0 +1,134 @@
...
+                      *.platform=linux/amd64,linux/arm64
+                  files: |
+                      docker-bake.hcl
+                      build.json
+                      docker-bake-template-meta.json
+
+            - name: Set output variables
+              id: bake_metadata
+              env:
+                  BAKE_META: ${{ steps.build-upload.outputs.metadata }}
```
danielhollas, line 132:
```suggestion
                  BAKE_METADATA: ${{ steps.build-upload.outputs.metadata }}
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,134 @@
...
+        outputs:
+            images:
+                description: Built images, identified with digests
+                value: ${{ jobs.build.outputs.images }}
+
+jobs:
+    build:
+        name: amd64 and arm64
+        runs-on: ${{ inputs.runsOn }}
+        timeout-minutes: 30
```
danielhollas, line 27:
```suggestion
        timeout-minutes: 60
```

danielhollas:
@unkcpz this is now ready from my side. I've written up the details at the top of the PR.

danielhollas:
> This is the initial reason that I want to separated the ARM64 build out to run on the ARM machine (and the complicity came from there, build arm64 on a separate machine require to manage/merge the docker tags after instead of using the bake-action). It was sometime not just take long time but the build would just fail. I didn't figure out what caused this but this will become a potential issue in the future.

Yeah, I am happy to do that in a separate PR.

unkcpz:
> Yeah, I am happy to do that in a separate PR.

Could you try to branch out from this PR and give it a small try, I personally don't think that fit well with using docker-bake metada action.

unkcpz:
> I believe some of the improvements here should be ported to QEApp and aiida-core repos as well.

For qeapp and aiida-core, the issues are not so stringent. Since all images are [build once and then tested for each of them](https://github.com/aiidateam/aiida-core/blob/main/.github/workflows/docker-build-test-upload.yml). It is easy to understand for QeApp since we only build one image. For aiida-core, we build three images on the same machine and tested on the same machine right after the build. 

Overall, I just a bit not so satisfied with using `merge-bake-template-target.sh` to prepare the metadata. To me this is not very standard way of using metadata-action, I took quite sometime to understand how it can manage preparing tags for multiple images. But I have to admit using artifact plus bake actions as I implemented is not good as you pointed out.

danielhollas:
> Overall, I just a bit not so satisfied with using merge-bake-template-target.sh to prepare the metadata. To me this is not very standard way of using metadata-action, I took quite sometime to understand how it can manage preparing tags for multiple images. But I have to admit using artifact plus bake actions as I implemented is not good as you pointed out.

That's a very fair point, I wasn't satisfied with it either. I've finally figured out how to completely get rid of it, please take a look at #446. I am really happy how it turned out.


> Could you try to branch out from this PR and give it a small try, I personally don't think that fit well with using docker-bake metada action.

I've opened #445 for exploration, and have gotten far enough to convince myself that it would be possible. In fact there's a big overlap with #446. The only thing missing there is the manifest creation and creating multiarch images, but should be doable using the approach we're currently using. So to me I'd really like to get #445 and this PR reviewed and merged, and then I can continue with #445.

stack/base-with-services/Dockerfile
```diff
@@ -10,45 +10,44 @@ ARG AIIDA_VERSION
...
+         rabbitmq-server=3.8.14  && \
      mamba clean --all -f -y && \
      fix-permissions "${CONDA_DIR}" && \
      fix-permissions "/home/${NB_USER}"; \
 elif [ "$TARGETARCH" = "arm64" ]; then \
-     apt-get update && apt-get install -y --no-install-recommends  \
-          erlang && \
-     rm -rf /var/lib/apt/lists/* && \
-     apt-get clean all && \
+     mamba install -p /opt/conda/envs/aiida-core-services --yes \
```
danielhollas, line 30:
```suggestion
     mamba create -p /opt/conda/envs/aiida-core-services --yes \
```

stack/base-with-services/Dockerfile
```diff
@@ -10,45 +10,44 @@ ARG AIIDA_VERSION
...
+#
+# RabbitMQ is not available on conda-forge at the time being, see:
+# https://github.com/conda-forge/rabbitmq-server-feedstock/issues/67If
+# Instead we need install erlang via apt and RabbitMQ as a "Generic Unix Build", see:
+# https://www.rabbitmq.com/install-generic-unix.html
+# Note that this version must be compatible with system's erlang version.
+# Note that system erlang from arm64 is already installed in the base image,
+# together with other APT dependencies to save build time.
 RUN if [ "$TARGETARCH" = "amd64" ]; then \
      mamba install -p /opt/conda/envs/aiida-core-services --yes \
```
danielhollas, line 23:
```suggestion
     mamba create -p /opt/conda/envs/aiida-core-services --yes \
```

stack/base-with-services/Dockerfile
```diff
@@ -10,45 +10,44 @@ ARG AIIDA_VERSION
...
      mamba clean --all -f -y && \
      fix-permissions "${CONDA_DIR}" && \
      fix-permissions "/home/${NB_USER}"; \
 elif [ "$TARGETARCH" = "arm64" ]; then \
-     apt-get update && apt-get install -y --no-install-recommends  \
-          erlang && \
-     rm -rf /var/lib/apt/lists/* && \
-     apt-get clean all && \
+     mamba create -p /opt/conda/envs/aiida-core-services --yes \
+         postgresql=${PGSQL_VERSION} \
```
danielhollas, line 31:
```suggestion
         postgresql=${PGSQL_VERSION} && \
```

danielhollas:
Last change, I've greatly simplified the build workflow, of rebuilding the amd64 images twice, but that only takes three minutes, and we can get them back later if we switch back to building on self-hosted runner anyway. The upside is, this PR now has -150 lines of YAML compared to main. :tada:

danielhollas:
@unkcpz will you have time to review this sometimes this week? I'd really like to finish this work since I spent quite some time on it. If you don't have time that's okay, just let me know. :-)

unkcpz:
Hi @danielhollas, I already start review it but not finish, I need to look into some details and look it as whole thing and test it locally a bit as well. 
Thanks for refactoring this, has a feeling that it is more clear. I'll give the feedback next week.

unkcpz:
>  a minute by installing erlang into base image

I think this is not correct, the `base` is for "not include any services". The services in the context of AiiDA is RMQ and postgresql. The erlang is specific for RMQ. 
In our case, the problem is we not actually using `base` and `base-with-services`. There are following options:
- Not publish `base` and `base-with-services`.
- Revert this change but requires 1 mins loss in building. 
- The `base` image is a little bit larger (usually it is you more picky about the image size :grin: ) but I am fine with it.

danielhollas:
> I think this is not correct, the base is for "not include any services". The services in the context of AiiDA is RMQ and postgresql. The erlang is specific for RMQ.

You are absolutely right, I have a comment about this in the `base` image Dockerfile.

https://github.com/aiidalab/aiidalab-docker-stack/pull/439/files#diff-2c89eac94aab2b39a07d7baffa58137a3fe3745c4698fd30483acb1906273b04R15

My thinking was that making base image a bit larger doesn't matter much (at least for now)  since we're currently not really using it, especially since you added Docker image to the aiida-core repo.
And it also makes full-stack Dockerfile simpler and centralizes all apt install into base. This might actually help to improve overall size of images that you have locally if you have different versions, because it might make it more likely that you will reuse layers from the base image.

unkcpz:
Hi @danielhollas, in the summary, this is not clear to me why that happened. There are for release jobs running from full-stack, are those for for different image respectively? Is that possible to give a label in the job name? 

image

.github/workflows/main.yml
```diff
@@ -0,0 +1,92 @@
...
+            images: ${{ needs.build-amd64.outputs.images }}
+            target: ${{ matrix.target }}
+            integration: false
+
+    build:
+        needs: test-amd64
+        uses: ./.github/workflows/build.yml
+        with:
+            runsOn: ubuntu-22.04
+            platforms: linux/amd64,linux/arm64
```
unkcpz, line 48: Is this means it build twice for `linux/amd64`?

docker-bake.hcl
```diff
@@ -40,10 +39,17 @@ variable "TARGETS" {
 function "tags" {
   params = [image]
   result = [
-    "${REGISTRY}${ORGANIZATION}/${image}:newly-build",
+    "${REGISTRY}${ORGANIZATION}/${image}${VERSION}",
```
unkcpz, line 42:
```suggestion
    "${REGISTRY}${ORGANIZATION}/${image}:${VERSION}",
```

Should this be with a `:`?  I am also not sure we still need the "<image>-meta" target after the refactoring. I think maybe the `tags` function can directly put into the `<image>` target?

.github/workflows/build.yml
```diff
@@ -0,0 +1,77 @@
...
+                  # Using provenance to disable default attestation so it will build only desired images:
+                  # https://github.com/orgs/community/discussions/45969
+                  provenance: false
+                  set: |
+                      *.platform=${{ inputs.platforms }}
+                      *.output=type=registry,push-by-digest=true,name-canonical=true
+                  files: |
+                      docker-bake.hcl
+                      build.json
+                      .github/workflows/env.hcl
```
unkcpz, line 70:
```suggestion
            set: |
                *.platform=linux/amd64
                *.output=type=registry,push-by-digest=true,name-canonical=true
                *.cache-from=type=local,src=/tmp/.buildx-cache
                *.cache-to=type=local,mode=max,dest=/tmp/.buildx-cache-new
            files: |
                docker-bake.hcl
                build.json

      # Temp fix for issue of fast size grow in local caching:
      # https://github.com/docker/build-push-action/issues/252
      # https://github.com/moby/buildkit/issues/1896
      - name: Move cache
        run: |
          rm -rf /tmp/.buildx-cache
          mv /tmp/.buildx-cache-new /tmp/.buildx-cache
```

Could you give it a try on caching? This may further speed the build?

docker-bake.hcl
unkcpz: And a nitpick.

`${REGISTRY}${ORGANIZATION}` -> `${REGISTRY}/${ORGANIZATION}`

with using `ghcr.io` instead of `ghcr.io/` in `env.hcl` probably more clear?

.github/workflows/build.yml
```diff
@@ -0,0 +1,77 @@
...
+              id: build-upload
+              uses: docker/bake-action@v4
+              with:
+                  push: true
+                  # Using provenance to disable default attestation so it will build only desired images:
+                  # https://github.com/orgs/community/discussions/45969
+                  provenance: false
+                  set: |
+                      *.platform=${{ inputs.platforms }}
+                      *.output=type=registry,push-by-digest=true,name-canonical=true
```
unkcpz, line 66: I don't understand why the `push-by-digest` is used for? From https://docs.docker.com/build/exporters/image-registry/ it says push without name. I assume with using this option, in the build phase the image should not tagged (see also my comment on `<image>-meta`, which give the tag for the image.)

.github/workflows/build.yml
danielhollas: Oh, interesting find, thanks! I would like to keep this PR as simple as possible so that we have a simple reference that we can always go back to.
I'd like to try it in a subsequent PR I that is okay with you.

docker-bake.hcl
danielhollas: Good catch. I actually tried both of those things and ran into issues. The way this currently works in CI is that the $image already includes both the registry and VERSION so I cannot include the separators here. Maybe something to improve in the future, but I didn't find an obvious way to do it.

danielhollas:
> Hi @danielhollas, in the summary, this is not clear to me why that happened. There are for release jobs running from full-stack, are those for for different image respectively? Is that possible to give a label in the job name?

Good point, I will tweak the names. Yes, those jobs are for the different images. The workflow is much simpler when they are pushed independently.

.github/workflows/build.yml
danielhollas, line 66:
```suggestion
                      *.output=type=registry,name-canonical=true
```

docker-bake.hcl
unkcpz: I have a feeling that the `VERSION` is not actually used, right?

danielhollas: I think it is still use when you run the build locally.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,94 @@
...
+              id: images
+              run: |
+                  src=$(echo '${{ inputs.images }}'| jq -cr '.[("${{ matrix.target }}"|ascii_upcase|sub("-"; "_"; "g")) + "_IMAGE"]')
+                  echo "src=$src" | tee -a "${GITHUB_OUTPUT}"
+
+            - name: Push image
+              uses: akhilerm/tag-push-action@v2.2.0
+              with:
+                  src: ${{ steps.images.outputs.src }}
+                  dst: ${{ steps.meta.outputs.tags }}
```
unkcpz, line 85: Hi @danielhollas, I am not sure with how to use this action but seems this might has some problem. The `tag-push-action` is only change the tag which require the image is already exist on registry. But it is not the case for docker.io. 

But maybe that will be more clear after merged and trigger the publish CI.

.github/workflows/extract-image-names.sh
```diff
@@ -0,0 +1,57 @@
...
+#
+# images={
+#   "BASE_IMAGE": "ghcr.io/aiidalab/base@sha256:8e57a52b924b67567314b8ed3c968859cad99ea13521e60bbef40457e16f391d",
+#   "BASE_WITH_SERVICES_IMAGE": "ghcr.io/aiidalab/base-with-services@sha256:6753a809b5b2675bf4c22408e07c1df155907a465b33c369ef93ebcb1c4fec26",
+#   "FULL_STACK_IMAGE": "ghcr.io/aiidalab/full-stack@sha256:85ee91f61be1ea601591c785db038e5899d68d5fb89e07d66d9efbe8f352ee48",
+#   "LAB_IMAGE": "ghcr.io/aiidalab/lab@sha256:4d9be090da287fcdf2d4658bb82f78bad791ccd15dac9af594fb8306abe47e97"
+# }
+#
+# This json output is later turned to environment variables using fromJson() GHA builtin
+# (e.g. BASE_IMAGE=ghcr.io/aiidalab/base@sha256:8e57a52b...)
```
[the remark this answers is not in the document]
danielhollas: I think it is correct, it is not a tag really, it's a digest.

.github/workflows/build.yml
danielhollas: I checked again, and it is needed, because without it a `latest` tag will be automatically added since we didn't specify any tags. See e.g. here: https://github.com/aiidalab/aiidalab-docker-stack/pkgs/container/base

image

This is a quite important trick that makes everything work by using digests instead of tags.

.github/workflows/publish.yml
danielhollas: > tag-push-action is only change the tag which require the image is already exist on registry.

Why do you think this? I am pretty sure this should work because we've been using this action precisely for this purpose in this repo. But we'll see after merge.

unkcpz: Previously we use tag, but from its example it didn't provide example for digest. But let's see after it merged.

.github/workflows/extract-image-names.sh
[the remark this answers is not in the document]
unkcpz: I see, make sense.

.github/workflows/publish.yml
danielhollas: Hmm, you might be right, let's see.

danielhollas:
@unkcpz I've pushed a fix to `dodo.py` for building locally. Can you try it out by running `doit build`? Here's what I get after that

```console
$ docker images
aiidalab/full-stack              2024.1017.post145.dev0_8f8a30e   1b069e00e91e   About a minute ago   2.54GB
aiidalab/lab                     2024.1017.post145.dev0_8f8a30e   af94ef50d9cc   About a minute ago   2.19GB
aiidalab/base-with-services      2024.1017.post145.dev0_8f8a30e   e6be8727a8d1   4 minutes ago        2.54GB
aiidalab/base                    2024.1017.post145.dev0_8f8a30e   dba20a49b4f7   34 minutes ago       2.16GB
```

The image tag (determined by VERSION variable) is automatically generated from git commit.

Otherwise I think I replied to all your comments, thank you for a detailed review! Once I get a green light from you I will merge and then make a release for the aiida v2.5 image.

unkcpz:
> Can you try it out by running doit build?

Thanks for the fix. This works for me locally.

But the command README seems not work `doit build --arch=arm64 --version=my-version`. And I specify with `--targe base` but it build all.

unkcpz:
Before merge this can we make a release with aiida-core==2.5.1 first?

danielhollas:
> But the command README seems not work doit build --arch=arm64 --version=my-version. And I specify with --targe base but it build all. 

This should now be fixed, can you try again? Note that the `doit tests` command is still broken, but this will require a bit more change to fix, I'll open an issue.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,93 @@
...
+              id: meta
+              uses: docker/metadata-action@v5
+              env: ${{ fromJSON(steps.build_vars.outputs.vars) }}
+              with:
+                  # e.g. ghcr.io/aiidalab/full-stack
+                  images: ${{ inputs.registry }}/${{ github.repository_owner }}/${{ matrix.target }}
+                  tags: |
+                      type=ref,event=pr
+                      type=edge,enable={{is_default_branch}}
+                      type=raw,value={{tag}},enable=${{ github.ref_type == 'tag' && ! startsWith(github.ref_name, 'v') }}
```
unkcpz, line 68:
```suggestion
                      type=raw,value={{tag}},enable=${{ github.ref_type == 'tag' && startsWith(github.ref_name, 'v') }}
```

This was added by accident I guess ;)

danielhollas: Nope, this is intentional. This was the original design, allowing to make a special tagged release by using a tag that doesn't start with `v`. This is also mentioned in the README. I am not too convinced that we need it, but I'd remove it in a separate PR, together with the documentation change.

unkcpz: I'd suggest to use the tag, I added this in the first place since I want to get both `2024.1001` and `v2024.1001` as tag. Because you mean don't remember whether the `v` need to be used when running `docker` command.

danielhollas: Fair enough, I've removed the line.

danielhollas:
I've temporarily switched to buildjet runners (and you can see that the tests pass now @unkcpz).
My plan is to merge this and try making a release again so we don't have to wait till the ARM64 runners are fixed.

.github/workflows/publish.yml
unkcpz: My suggestion was to keep the tag like `v2024.1001`. Since it is a tag that in use such as 
image

I use it in PSI server deployment. I can make the change but for users which using `v2024.1001` may confused if we remove it.

unkcpz:
> My plan is to merge this and try making a release again so we don't have to wait till the ARM64 runners are fixed.

Make sense to me. And I think we really need to move to buildjet.

danielhollas:
@unkcpz can you approve so I can merge? (I could bypass it but don't want to :-) )

unkcpz (approved):
All good. This is much clear! Thanks. @danielhollas (and I know how difficult to figure out all the details 😄 )

.github/workflows/publish.yml
danielhollas: It worked!
https://github.com/aiidalab/aiidalab-docker-stack/actions/runs/9022411878/job/24792606455

.github/workflows/publish.yml
danielhollas: Yes, but that is already handled in the match group below I believe.

PR #440, opened 2024-03-27
Pin aiida-core version

### LINKED ISSUE #436: Pin the aiida-core to its minor version
Currently, we don't allow the user to downgrade the aiida-core version below what's in the image, since it might not be compatible due to DB migration. However, we currently _do allow_ users to upgrade the minor version of AiiDA.

Here I would propose to disallow that, and have aiida-core pinned (whether to pin up to the patch version I am not sure), in the hopes of making the AiiDAlab container more stable. Here are some concrete reasons / scenarious:

1. aiida-core brings in a lot of dependencies, so there's a greater chance of things breaking. Indeed, it has been so far our experience that upgrading the minor version often necessitated changes throughout the AiiDAlab stack. As a concrete example, upgrading to v2.5 would cause breakage to any packages that depend on pydantic v1 (`optimade`, `aiida-quantumespresso`).

2. Upgrading aiida-core possibly involved unreversible DB migration. So if something goes wrong, the user cannot go back to previous aiida-core version, resulting in a completely broken state.

3. Reinstalling aiida-core puts the package in `~/.local/`, and mixes conda and pip dependencies, which I think is undesirable, because if the user then upgrades the image with a newer aiida-core version, there will be a conflict.

Given that we tag all AiiDAlab images with the AIIDA-CORE version, I think it is completely reasonable to require users to download a new image if they want to upgrade aiida-core.

CC @yakutovicha @unkcpz what do you think? (if you agree, it would be good to make this change before we update aiida-core to 2.5)

danielhollas:
Closes #436

### DIFF
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -16,11 +16,12 @@ WORKDIR /opt/
 
 ARG AIIDA_VERSION
 
-# Pin shared requirements in the base environemnt.
-# The lower bound of aiida-core is set to the version of aiida-core to prevent
-# lower aiida-core DB version installed and trigger database downgrade.
+# Pin shared requirements in the base environment.
+# We pin aiida-core to the exact installed version,
+# to prevent accidental upgrade or downgrade, that might
+# induce DB migration or break shared dependencies of AiiDAlab Apps.
 RUN echo "pip==23.3.1" > /opt/requirements.txt && \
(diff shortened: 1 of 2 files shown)


danielhollas:
@unkcpz thanks for fixing the arm runner, this should now be ready.

yakutovicha (approved):
Go ahead @danielhollas, thanks a lot 👍

PR #442, opened 2024-04-08
Refactor tests

danielhollas:
Currently, one has to manually pass in test files for tests for various image variants (targets). That's not very elegant; instead here we automatically skip all tests that are not intended to pass for a given image.
Note that there is a behavioural change: Previously, for e.g. `lab` variant, we would only run `test-common.py` and `test-lab.py`. But now we also test it against `test-base.py`. All images should pass all the tests for the lower layers as well. `full-stack` should pass all tests.

Also switched from black to ruff and added basic lint rules to help with writing tests.


danielhollas:
@unkcpz this one is ready for review, LMK if you'll have a time to have a look (or perhaps nominate another reviewer?)

unkcpz (approved):
Thanks @danielhollas, changes are all clear and looks all good.

PR #443, opened 2024-04-17
Add comment on the app integration test fixture

unkcpz:


danielhollas:
Thanks, I'll incorporate the note as part of #439.
Unfortunately, I don't think there is anything actionable here since this is essentially a limitation of pip.

PR #444, opened 2024-04-17
WIP: Try setup-python on arm runners

danielhollas:
According to https://github.com/actions/setup-python/issues/108#issuecomment-1589663385 this might work now? Let's test!

Might solve #432

### DIFF
--- a/.github/actions/create-dev-env/action.yml
+++ b/.github/actions/create-dev-env/action.yml
@@ -12,7 +12,7 @@ runs:
     using: composite
     steps:
         - name: Set Up Python 🐍
-          uses: actions/setup-python@v4.0.0
+          uses: actions/setup-python@v4.1.0
           with:
               python-version: 3.11
               cache: pip
--- a/.github/actions/create-dev-env/action.yml
+++ b/.github/actions/create-dev-env/action.yml
@@ -12,7 +12,7 @@ runs:
     using: composite
     steps:
         - name: Set Up Python 🐍
-          uses: actions/setup-python@v4.0.0
+          uses: actions/setup-python@v5
           with:
               python-version: 3.11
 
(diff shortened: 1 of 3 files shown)


danielhollas:
setup-python@v5 gives a permission error `/Users/runner: Permission denied`, same as described in https://github.com/actions/setup-python/issues/792. This unmerged PR might fix it: https://github.com/actions/setup-python/pull/708

danielhollas:
With version `4.0.0` it seems I am getting a bit further, but failing right after creating hostedtoolcache folder:

```
Check if Python hostedtoolcache folder exist...
Creating Python hostedtoolcache folder...
Install Python binaries from prebuilt package
Error: sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
Error: sudo: a password is required
Error: The process '/bin/bash' failed with exit code 1
```

danielhollas:
@mbercx @unkcpz I think in order to make progress here we'll need tweak the runner machine per README instructions for `setup-python` action: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#macos

Specifically:

> 
    Create a directory called /Users/runner/hostedtoolcache
    Change the permissions of /Users/runner/hostedtoolcache so that the runner has write access

I think this is of reasonably high priority since right now we seem to be using a global python environment for all the jobs? See #432

mbercx:
Thanks @danielhollas! I'll admit it's mostly @unkcpz who is the expert on setting up/configuring the runner. I'm just the guy who owns the work station we're using for it. 😅 

@unkcpz I've packed my mac Mini and will bring it to the office tomorrow. Then we can consider using that instead, since I often use a VPN on my work station (or need to reboot, rarely), which interrupts the runner service.

danielhollas:
This PR is now stale, but the issue still needs to be solved, unless we move to buildjet runners (I am slightly worried it would be expensive though, since I am running through my free 5$ credit quite quickly).

PR #445, opened 2024-04-23
WIP: Build on arm64

danielhollas:
Builds on top of #439


unkcpz:
Hi @danielhollas, since the self-hosted runner machine not stable. I recommend to use https://buildjet.com/for-github-actions to check the arm64 bulid. It provide 3 months free trail, sadly I had run out of it last year for support arm64 test. It should be simple by just switch the `runs-on` tag in workflow after activate the account.

danielhollas:
> Hi @danielhollas, since the self-hosted runner machine not stable. I recommend to use https://buildjet.com/for-github-actions to check the arm64 bulid. It provide 3 months free trail, sadly I had run out of it last year for support arm64 test. It should be simple by just switch the runs-on tag in workflow after activate the account.

Thanks, I will try that. Will you have time to look with Marnik on the ARM64 runner? For one, I really need jq installed there.

unkcpz:
> Thanks, I will try that. Will you have time to look with Marnik on the ARM64 runner? For one, I really need jq installed there.

Sure, I thought he already did that, since he ask me last week. I'll check with him now.

danielhollas:
I will not be able to finish this today, but essentially we will need to implement this approach from official docker docs:

https://docs.docker.com/build/ci/github-actions/multi-platform/#with-bake

danielhollas:
Most of this PR has been done as part of #446, closing for now. If / when we need to switch building on self-hosted runner I'll open a new one.

PR #446, opened 2024-04-23
Remove bake template

danielhollas:
This PR builds on top of #439, I've pulled it out to a separate PR for easier review. 

This PR gets rid of the rather obnoxious merge-bake-template-target.sh script.
 
The trick here is to NOT use the `docker/metadata-action` when first building the image:
we can simply not tag the image and use image digest for identifying it. This is in any case preferred since the digest is an SHA256 hash that uniquely identifies the image, unlike a tag that can move at any time.

Only after all the images are built and tested, we tag them and publish them to ghcr.io, using the same workflow that we already have for publishing to docker.io
Unlike in build step, where we build all the images together, the push can happen for each image separately so we don't need to do this weird bake-template dance.

Here's how the workflow looks in this PR. (note that here we're building both amd64 and arm64 together on GHA ubuntu runner.)

image

@unkcpz if this looks good to you, I will merge this into #439 and we can continue review there.


unkcpz:
Super cool! I spend couple of hours and didn't figure it out, you beat me ;)  I'll have a look at ASAP.

.github/workflows/build.yml
```diff
@@ -48,54 +44,36 @@ jobs:
...
                       build.json
-                      docker-bake-template-meta.json
+                      .github/workflows/env.hcl
 
             - name: Get amd64 images with digests
               id: bake_metadata_amd64
               env:
                   BAKE_METADATA: ${{ steps.build-amd64.outputs.metadata }}
               run: |
+                  .github/workflows/extract-image-names.sh
```
unkcpz, line 72: Why this need to be run twice?

.github/workflows/publish.yml
```diff
@@ -53,16 +62,18 @@ jobs:
               uses: docker/metadata-action@v5
               env: ${{ fromJSON(steps.build_vars.outputs.vars) }}
               with:
-                  images: docker.io/aiidalab/${{ matrix.target }}
+                  # e.g. ghcr.io/aiidalab/full-stack
+                  images: ${{ inputs.registry }}/${{ github.repository_owner }}/${{ matrix.target }}
                   tags: |
+                      type=ref,event=pr
```
unkcpz, line 68: If I understand correctly, the `release` workflow will triggered not only by the tag but also for every PR and this is why you include it here to have ghcr.io to have PR tag as we had before.

.github/workflows/main.yml
```diff
@@ -16,6 +16,9 @@ concurrency:
     group: ${{ github.workflow }}-${{ github.ref }}
     cancel-in-progress: true
 
+env:
+    FORCE_COLOR: 1
```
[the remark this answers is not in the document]
danielhollas: This is a trick I learned recently that forces various CLI tools to print colors, even though they are not run in tty mode. Thanks to this, Github logs should be more colorful! :rainbow:  See https://force-color.org/ for more explanation.

.github/workflows/publish.yml
danielhollas: Yes, I am reusing this workflow to publish images both to ghcr.io and docker.io. The former runs always, while the latter runs only on release or push to main. See the main.yml file.

unkcpz (approved):
Changes looks all good, let's merge to `remove-artifacts` and I'll do more review there. Thanks!

PR #453, opened 2024-05-09
Reduce arm usage

danielhollas:
I am already out of my free credits on buildjet. Here I am changing the workflow so that the ARM tests only run on main branch. It's not ideal but currently that's the easiest way forward.
I also try to use caching, it seems to help a little bit, but not fully (e.g. the arm part of the build is not cached). We can try to tweak it in a follow up PR.


danielhollas:
@unkcpz can you take a look? I'll try to make a release after merging this (though I am afraid that I've run out of free buildjet credit, let's see).

unkcpz (approved):
Thanks @danielhollas, the caching seems working and all changes looks good.

PR #456, opened 2024-05-16
Speedup container startup

danielhollas:
The main strategy here is to remove all unneeded verdi invocations, because each one costs at least 0.5s.

### DIFF
--- a/stack/base-with-services/before-notebook.d/20_start-postgresql.sh
+++ b/stack/base-with-services/before-notebook.d/20_start-postgresql.sh
@@ -1,9 +1,6 @@
 #!/bin/bash
 set -x
 
-# Activate the conda environment with PostgreSQL installed in it.
-# conda activate pgsql
-
 # -w waits until server is up
 PSQL_START_CMD="pg_ctl --timeout=180 -w -D /home/${NB_USER}/.postgresql -l /home/${NB_USER}/.postgresql/logfile start"
 PSQL_STOP_CMD="pg_ctl -w -D /home/${NB_USER}/.postgresql stop"
@@ -18,19 +15,15 @@ if [ ! -d /home/${NB_USER}/.postgresql ]; then
    echo "unix_socket_directories = '/tmp'" >> /home/${NB_USER}/.postgresql/postgresql.conf
    ${MAMBA_RUN} ${PSQL_START_CMD}
 
-# else don't
 else
     # Fix problem with kubernetes cluster that adds rws permissions to the group
     # for more details see: https://github.com/materialscloud-org/aiidalab-z2jh-eosc/issues/5
     chmod g-rwxs /home/${NB_USER}/.postgresql -R
 
-    # stores return value in $?
-    running=true
-    ${MAMBA_RUN} ${PSQL_STATUS_CMD} || running=false
-
-    # Postgresql was probably not shutdown properly. Cleaning up the mess...
-    if ! $running ; then
-       echo "" > /home/${NB_USER}/.postgresql/logfile # empty log files
+    if ! ${MAMBA_RUN} ${PSQL_STATUS_CMD}; then
+       # Cleaning up the mess if Postgresql was not shutdown properly.
+       # TODO: Rotate logfile
+       echo "" > /home/${NB_USER}/.postgresql/logfile
        rm -vf /home/${NB_USER}/.postgresql/postmaster.pid
        ${MAMBA_RUN} ${PSQL_START_CMD}
    fi
--- a/stack/base-with-services/before-notebook.d/20_start-postgresql.sh
+++ b/stack/base-with-services/before-notebook.d/20_start-postgresql.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -x
 
 # Activate the conda environment with PostgreSQL installed in it.
 # conda activate pgsql
--- a/stack/base/before-notebook.d/40_prepare-aiida.sh
+++ b/stack/base/before-notebook.d/40_prepare-aiida.sh
@@ -1,15 +1,12 @@
 #!/bin/bash
 
 # This script is executed whenever the docker container is (re)started.
-
-# Debugging.
 set -x
 
-# Environment.
 export SHELL=/bin/bash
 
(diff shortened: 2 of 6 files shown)


danielhollas:
@unkcpz can you test if this image from this PR starts faster? (I would expect at least 5s faster, not solving your problem, but still nice)

unkcpz:
Thanks @danielhollas, I also notice the slow down of the unnecessary migration I think this is the one bring most improvement. Will give this a look!

danielhollas:
On my machine, I ran 

```
time podman run --rm ghcr.io/aiidalab/full-stack:pr-456
```

and it took ~13s, while the latest image takes ~18s.

stack/base/before-notebook.d/40_prepare-aiida.sh
```diff
@@ -48,7 +48,8 @@ if [[ ${NEED_SETUP_PROFILE} == true ]]; then
       exit 1
     fi
 
-    verdi computer show ${computer_name} || verdi computer setup \
+    # TODO: Load config from YAML file
```
unkcpz, line 51: This is not urgent and didn't bring speedup I assume? Maybe using verdi is more straightforward.

unkcpz (changes requested):
Hi @danielhollas, let me know when this is ready for review.

stack/base/before-notebook.d/40_prepare-aiida.sh
danielhollas: No, this wouldn't bring any speed up since you would still need to invoke verdi. It would be a bit cleaner since the computer would be statically defined in a yaml file, and not via cmdline parameters, but it's not important for this PR.

stack/base/before-notebook.d/40_prepare-aiida.sh
```diff
@@ -61,17 +62,15 @@ if [[ ${NEED_SETUP_PROFILE} == true ]]; then
...
 
-# Make sure that the daemon is not running, otherwise the migration will abort.
-verdi daemon stop
+  # Migration will run for the default profile.
+  verdi storage migrate --force
 
-# Migration will run for the default profile.
-verdi storage migrate --force
+  # Show the default profile
+  verdi profile show
```
danielhollas, line 72: We could in principle get rid of this, it's probably not important here.

stack/base/before-notebook.d/40_prepare-aiida.sh
```diff
@@ -61,17 +62,15 @@ if [[ ${NEED_SETUP_PROFILE} == true ]]; then
...
         --non-interactive                                               \
         --safe-interval 0.0
-fi
-
 
-# Show the default profile
-verdi profile show || echo "The default profile is not set."
+else
 
-# Make sure that the daemon is not running, otherwise the migration will abort.
```
danielhollas, line 70: I got rid of this, since by definition when this script is run at startup the daemon cannot be running. It does make this script a bit less robust against future changes elsewhere, but since we control the stack I think it's fine (and issues here would be caught by tests)

stack/base/before-notebook.d/40_prepare-aiida.sh
```diff
@@ -48,7 +45,7 @@ if [[ ${NEED_SETUP_PROFILE} == true ]]; then
       exit 1
     fi
 
-    verdi computer show ${computer_name} || verdi computer setup \
```
danielhollas, line 51: Again, since we have just created a profile, by definition there can't be any computer yet!

stack/base/before-notebook.d/40_prepare-aiida.sh
```diff
@@ -1,15 +1,12 @@
...
 # This script is executed whenever the docker container is (re)started.
-
-# Debugging.
 set -x
 
-# Environment.
 export SHELL=/bin/bash
 
 # Check if user requested to set up AiiDA profile (and if it exists already)
-if [[ ${SETUP_DEFAULT_AIIDA_PROFILE} == true ]] && ! verdi profile show ${AIIDA_PROFILE_NAME} &> /dev/null; then
```
danielhollas, line 12: If the profile already exists, it will get printed here (and we don't need to print it again below)

stack/base/before-notebook.d/40_prepare-aiida.sh
danielhollas: On second thought, since the number of CPUs is defined dynamically, I think we can't have config in yaml. I've removed the TODO.

stack/base/before-notebook.d/40_prepare-aiida.sh
danielhollas: I got rid of it, we will now print the profile at the top if it exist.

danielhollas:
@unkcpz ready for review. I got rid of one more verdi invocation.

danielhollas:
@superstar54 this one is ready to go. Would be great if you could have a look when you have a bit of time. Thanks!

superstar54 (approved):
LGTM!

stack/base-with-services/before-notebook.d/20_start-postgresql.sh
```diff
@@ -17,19 +15,15 @@ if [ ! -d /home/${NB_USER}/.postgresql ]; then
...
-    # stores return value in $?
-    running=true
-    ${MAMBA_RUN} ${PSQL_STATUS_CMD} || running=false
-
-    # Postgresql was probably not shutdown properly. Cleaning up the mess...
-    if ! $running ; then
-       echo "" > /home/${NB_USER}/.postgresql/logfile # empty log files
+    if ! ${MAMBA_RUN} ${PSQL_STATUS_CMD}; then
+       # Cleaning up the mess if Postgresql was not shutdown properly.
+       # TODO: Rotate logfile
```
unkcpz, line 25: Hi @danielhollas, what is this means, do you gonna to address this before it get merged?

danielhollas: I've actually addressed this in #457, have a look.

unkcpz: Make sense.

PR #457, opened 2024-05-21
Initialize DB during image build

danielhollas:
Speed up container startup by initialized Posgresql DB during image build. 

Turns out we were already doing that in the `base-with-services` image, but this initialized DB had a wrong name, inconsistent with `20_start-postgresql.sh`. To prevent this mismatch in the future, we now set `PGDATA` envvar in the image, which is automatically picked up by `initdb` and `pg_ctl` commands. See e.g. https://www.postgresql.org/docs/current/app-initdb.html#APP-INITDB-OPTION-PGDATA

This PR is build on top of #456. Taking these changes all together, the full-stack image now starts up in 9s on my machine.


stack/full-stack/Dockerfile
```diff
@@ -1,16 +1,20 @@
 # syntax=docker/dockerfile:1
 FROM base-with-services as base
+ARG PGDATA_FOLDER
```
danielhollas, line 3:
```suggestion
ARG PGDATA_FOLDER
FROM base-with-services as base
ARG PGDATA_FOLDER
```

stack/full-stack/Dockerfile
```diff
@@ -1,16 +1,21 @@
...
 USER root
 
 COPY --from=base /opt/config-quick-setup.yaml /opt/
 COPY --from=base "${CONDA_DIR}/envs/aiida-core-services" "${CONDA_DIR}/envs/aiida-core-services"
 COPY --from=base /usr/local/bin/before-notebook.d /usr/local/bin/before-notebook.d
+COPY --from=base ${PGDATA} ${PGDATA}
 
 RUN fix-permissions "${CONDA_DIR}"
-RUN fix-permissions "/home/${NB_USER}/.aiida"
+RUN fix-permissions "/home/${NB_USER}"
```
danielhollas, line 18:
```suggestion
RUN fix-permissions "/home/${NB_USER}/.aiida"
```

stack/full-stack/Dockerfile
```diff
@@ -1,13 +1,18 @@
 # syntax=docker/dockerfile:1
+ARG PGDATA_FOLDER
 FROM base-with-services as base
+ARG PGDATA_FOLDER
 
 FROM lab
 
+ENV PGDATA=${PGDATA_FOLDER}
```
danielhollas, line 8:
```suggestion
ARG PGDATA_FOLDER
ENV PGDATA=${PGDATA_FOLDER}
```

stack/base-with-services/before-notebook.d/20_start-postgresql.sh
```diff
@@ -2,29 +2,26 @@
...
-    # for more details see: https://github.com/materialscloud-org/aiidalab-z2jh-eosc/issues/5
-    chmod g-rwxs /home/${NB_USER}/.postgresql -R
+    chmod -R g-rwxs ${PGDATA}
 
     if ! ${MAMBA_RUN} ${PSQL_STATUS_CMD}; then
        # Cleaning up the mess if Postgresql was not shutdown properly.
-       # TODO: Rotate logfile
-       echo "" > /home/${NB_USER}/.postgresql/logfile
-       rm -vf /home/${NB_USER}/.postgresql/postmaster.pid
+       mv ${PGDATA}/logfile ${PGDATA/logfile.1 && gzip ${PGDATA}/logfile.1
```
danielhollas, line 23:
```suggestion
       mv ${PGDATA}/logfile ${PGDATA}/logfile.1 && gzip ${PGDATA}/logfile.1
```

docker-bake.hcl
```diff
@@ -54,22 +54,8 @@ group "default" {
   targets = "${TARGETS}"
 }
 
-target "base-meta" {
```
danielhollas, line 57: Unrelated cleanup

unkcpz:
Are you sure this is the correct way to go? The purpose of initialize DB during the runtime phase is to having the DB in home and therefore in the persistent volume.

danielhollas:
I am not sure what you mean @unkcpz, the DB here **is** initialized in the home directory. So this should work when using aiidalab-launch and normal Docker. In Kubernetes, you'll have the same issue as with everything else in the home directory.

unkcpz:
> In Kubernetes, you'll have the same issue as with everything else in the home directory.

For everything (not everything but something) else, we do "home prepare" with https://github.com/aiidalab/aiidalab-docker-stack/blob/main/stack/base/before-notebook.d/10_prepare-home-config.sh. This not happened to the DB folder. 
Either we adapt the same strategy as we are planing for the qeapp image to have a `home` some where else and load it back after. Otherwise, with your changes here, the `full-stack` will fail on k8s deployment.

danielhollas:
> For everything (not everything but something) else, we do "home prepare" with https://github.com/aiidalab/aiidalab-docker-stack/blob/main/stack/base/before-notebook.d/10_prepare-home-config.sh. This not happened to the DB folder.

Yes, but that's because of k8s right?

> Either we adapt the same strategy as we are planing for the qeapp image to have a home some where else and load it back after. Otherwise, with your changes here, the full-stack will fail on k8s deployment.

I don't think so, I didn't actually changed the logic in `stack/base-with-services/before-notebook.d/20_start-postgresql.sh`. If the DB folder doesn't exist, it will get created at startup, as was done till now, so k8s should work I believe.

unkcpz:
>  I didn't actually changed the logic in stack/base-with-services/before-notebook.d/20_start-postgresql.sh

I see, I didn't take close look at the PR, sorry. Then all make sense.

.github/workflows/build.yml
```diff
@@ -74,6 +74,6 @@ jobs:
             - name: Set output variables
               id: bake_metadata
               run: |
-                  .github/workflows/extract-image-names.sh | tee -a "${GITHUB_OUTPUT}"
+                  .github/workflows/extract-image-names.sh | tee -a "${GITHUB_OUTPUT}" | awk -F'=' '{print $2}' | jq
```
unkcpz, line 77: This I don't understand, what different it will make?

stack/base-with-services/Dockerfile
```diff
@@ -58,4 +65,4 @@ USER ${NB_USER}
 WORKDIR "/home/${NB_USER}"
 
 # Initialize the database
-RUN mamba run -n aiida-core-services initdb -D aiida_db -U aiida
+RUN mamba run -n aiida-core-services initdb
```
unkcpz, line 68: What is the DB name used when create aiida profile after this change? We set it explicitly in `config-quick-setup.yaml` as `aiida_db`. 

BTW, we need keep in mind to adapt the change to use `verdi presto` after we move to 2.6.0

unkcpz (changes requested):
Thanks @danielhollas! Sorry for the late review. 

The changes are all good, just some minor questions. 

Since we are in the place, it is worth to discuss, do we really need a environment `aiida-core-services`? But for sure it will be another PR if we decide to move the conda env into `base`.

stack/base-with-services/Dockerfile
danielhollas: Good question. `-D` parameter doesn't set the db name, it sets the DB location.

https://www.postgresql.org/docs/current/app-initdb.html#APP-INITDB-OPTION-PGDATA

 in this case it would set it to `~/aiida_db`. But this was broken anyway, since the location in `stack/base-with-services/before-notebook.d/20_start-postgresql.sh` was set to `~/.postgresql`, so essentially this initdb command was being ignored. Now the location is set via the PGDATA variable and thus is guaranteed to be the same everywhere to prevent these kind of mismatches in the future.

Note that the the `-U` parameter sets the so-called [bootstrap superuser](https://www.postgresql.org/docs/current/glossary.html#GLOSSARY-BOOTSTRAP-SUPERUSER) I am not sure if this would have any effect, but we are not using this parameter in `stack/base-with-services/before-notebook.d/20_start-postgresql.sh` so I deleted it here as well.

danielhollas: > BTW, we need keep in mind to adapt the change to use verdi presto after we move to 2.6.0

I don't think we'll want to use `verdi presto` since we'd loose some config options. Instead I believe we should migrate from `verdi quicksetup` to the full `verdi profile setup`. But that's not that urgent, quicksetup has been deprecated but not removed.

.github/workflows/build.yml
danielhollas: Oh, this just pretty prints the json output so that it is easier to read in the Github Actions logs. Previously it was on a single line, and it was a bit annoying when I needed to copy the image name.

danielhollas:
> Since we are in the place, it is worth to discuss, do we really need a environment aiida-core-services? But for sure it will be another PR if we decide to move the conda env into base.

Why would you want to move it to the base environment? base environment contains literally everything else so I quite like that we can have at least these services in a separate one. If you disagree, could you please open an issue with your reasoning?

.github/workflows/build.yml
danielhollas: This is how it looks now in the amd64 image build

image

stack/base-with-services/Dockerfile
unkcpz: Makes sense, thanks!

unkcpz:
> Why would you want to move it to the base environment? base environment contains literally everything else so I quite like that we can have at least these services in a separate one.

The main reason is of this `mlocate` which is in the `aiida-core-services` environment and required to run `pytest`. It just over complex to have an extra conda environment that we gain little from it.

unkcpz (approved):
Thanks! All good to me. @danielhollas I'll let you merge and write up for the commit message.

PR #458, opened 2024-05-23
Fix doit tests

### LINKED ISSUE #451: Fix running tests with doit
`doit tests` currently does not work since the test suite was changed. We need to think about whether we want to use doit at all for tests, or whether to simply invoke pytest directly.

danielhollas:
Fixes #451. You can now build images locally with e..g

```console
doit build
```

and then run tests as 

```console
doit tests -t full-stack
```

I've also made some tweaks to support podman (and podman-compose) via extra `--compose-cmd` arguments. I've also fixed the `doit up` and `doit down` commands, which now have to include the `-t/--target` argument.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ The build system will attempt to detect the local architecture and automatically
 All commands `build`, `tests`, and `up` will use the locally detected platform and use a version tag based on the state of the local git repository.
 However, you can also specify a custom platform or version with the `--platform` and `--version` parameters, example: `doit build --arch=arm64 --version=my-version`.
 
-You can specify target image variants to build with `--target`, example: `doit build --target base --target lab`.
+By default, all image variants are build. You can specify a single target image variant to build with `-t/--target`, example: `doit build --target base`.
 
 ### Run automated tests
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -55,21 +55,21 @@ The build system will attempt to detect the local architecture and automatically
 All commands `build`, `tests`, and `up` will use the locally detected platform and use a version tag based on the state of the local git repository.
 However, you can also specify a custom platform or version with the `--platform` and `--version` parameters, example: `doit build --arch=arm64 --version=my-version`.
 
-You can specify target image variants to build with `--target`, example: `doit build --target base --target lab`.
```
danielhollas, line 58: To simplify the code, you can now only provide a single `--target` argument (and we now automatically validate that you choose one of the four valid images. If no target is specified, then all images are built by default.

dodo.py
```diff
@@ -15,23 +15,25 @@
 
 _ARCH_MAPPING = {
     "x86_64": "amd64",
+    "amd64": "amd64",
```
danielhollas, line 18: This is what podman returns on Fedora 39 at least.

danielhollas:
I've merged this since changes here don't influence the build / image itself, and we can always tweak later.

PR #461, opened 2024-05-26
Use one RabbitMQ version for both amd64 and arm64 builds

danielhollas:
I don't see much of a reason for using different RabbitMQ version for amd64 and arm64. 
Using the same version for both allows for nice simplification of the logic. The RMQ_VERSION variable is now defined in a single place in `build.json`, like all the others.

I've also moved rabbitmq configuration into its own script that is run at build time (instead of running it on every container startup).


.github/workflows/main.yml
```diff
@@ -58,9 +58,9 @@ jobs:
     # To save arm64 runner resources, we run the tests only on main
     # and only for full-stack image (same for integration tests below).
     test-arm64:
-        if: >-
-            github.repository == 'aiidalab/aiidalab-docker-stack'
-            && (github.ref_type == 'tag' || github.ref_name == 'main')
+        #if: >-
```
danielhollas, line 61: TODO: Revert this change, just temporary for testing.

unkcpz:
The main issue I had was not sure how to get the version of RMQ installed from system. It will install the version from which given by apt and it is not controlled by us. So instead of keeping the version are the same, using the `3.8.14` from conda for amd64 is the safe choose. We need to use different way to install (and setup) RMQ for amd64 and arm64 anyway. I didn't see the problem that the version are the same.

danielhollas:
I understand that currently you have to install rmq differently in arm64. But I still think it doesn't make much sense to use different versions. This is why I asked in aiida discourse if other RMQ versions work, and apparently there's no problem with higher versions besides the extra configuration, which we have under control anyway.

If you look at the actual changes, centralizing the RMQ version leads to a nice simplification in a couple of places. Also, I am hoping that soon we will have RabbitMQ available in conda for arm64, so this is a step in that direction.

danielhollas:
I rebased on main, the diff should now be clear, sorry about that. @unkcpz

stack/base-with-services/Dockerfile
```diff
@@ -29,36 +30,34 @@ ENV PGDATA=/home/${NB_USER}/.postgresql
...
+  fi
 
 # Configure AiiDA profile.
 COPY config-quick-setup.yaml .
 COPY before-notebook.d/20_start-postgresql.sh /usr/local/bin/before-notebook.d/
-COPY before-notebook.d/30_start-rabbitmq-${TARGETARCH}.sh /usr/local/bin/before-notebook.d/
+COPY before-notebook.d/30_start-rabbitmq.sh /usr/local/bin/before-notebook.d/
+
+# Configure RabbitMQ
+COPY configure_rabbitmq.sh .
```
danielhollas, line 54: NOTE: I've separated part of the startup script related to RMQ configuration and now it runs during image build and not at container startup. This makes startup a tiny bit faster, but the main benefit for me that the rmq startup script (`30_start-rabbitmq.sh`` is now very simple.

stack/base-with-services/before-notebook.d/30_start-rabbitmq.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -emx
+
+# Fix issue where the erlang cookie permissions are corrupted.
+chmod 400 "/home/${NB_USER}/.erlang.cookie" || echo "erlang cookie not created yet."
+
+mamba run -n aiida-core-services rabbitmq-server -detached
```
unkcpz, line 7: I don't understand, for arm64, the rabbitmq-server is not installed from conda. I guess this will fail for the arm64.
This script will run both for the amd64 and arm64 I assume. Originally, the script is different copied based on the arch.

danielhollas: Very good catch, not sure how I missed it, thanks. I did trigger an arm64 test originally but have removed those commits when rebasing. Will test again.

danielhollas: I think this might actually work, since in arm64 rabbitmq is available globally, so it doesn't matter whether it is run through conda environment or not.

danielhollas: The tests passed so I think this approach works. @unkpcz if you have a Mac you can test manually if you want. 

I added a note with an explanation. Hopefully soon we'll have rabbitmq conda package for arm64 as well

unkcpz: Okay, thanks for checking. All good for me then.

build.json
```diff
@@ -6,6 +6,9 @@
     "PGSQL_VERSION": {
       "default": "15"
     },
+    "RMQ_VERSION": {
+      "default": "3.9.13"
```
unkcpz, line 10: But this version is only for the RMQ from conda. It in principle not able to get the concrete version of RMQ from apt. I think this just bring us unnecessary maintain burden to align the version. 

But I am still fine with the refactoring, which is indeed more clear.

unkcpz (approved):
Thanks @danielhollas. See my last minor comment, it is good to go.

build.json
danielhollas: But we don't get RMQ via APT in arm64 builds, we download it directly from Github releases and can select which version we want:

https://github.com/aiidalab/aiidalab-docker-stack/blob/3208794c1bf1f6faf4de34c4baed8b3a5622b930/stack/base-with-services/Dockerfile#L41

We *do* install Erlang from apt and don't have control over its version, so we can only choose RMQ versions that are supported by it, that is true. There is a big comment about it in the Dockerfile linked above. I agree that is a problem. I am hoping that before the next version bump we'll have conda packages on arm64, I'll push for that at the end of this month.

PR #464, opened 2024-06-04
Increase the cull_inactive_interval for terminal app to 1 hour

danielhollas:
While testing the DEMO server @cpignedoli noticed that the Terminal app auto-closes relatively quickly, which might be annoying to users. In this PR I am increasing the "culling" interval from 10 minutes to 1 hour.


superstar54 (approved):
LGTM!

PR #465, opened 2024-06-10
Adding the installation of h5py==3.11.0 in the dockerfile

mikibonacci:
Only if architecture is arm64. Otherwise the aiidalab-qe-vibroscopy plugin will fail to install. The reason is that we cannot install easily h5py via pip, for arm64.

This should solve issue https://github.com/aiidalab/aiidalab-qe/issues/745 and https://github.com/mikibonacci/aiidalab-qe-vibroscopy/issues/52

### DIFF
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -15,15 +15,13 @@ ENV EXTRA_APT_PACKAGES "curl povray rsync build-essential"
 # For ARM64 we need to install erlang as it is not available on conda-forge
 # (this is needed later as rabbitmq dependency in base-with-services image,
 # but we install it here so that we don't have to invoke apt multiple times.
-# the h5py==3.11.0 is required in arm64 arch to install smoothly aiidalab-qe-vibroscopy.
 ARG TARGETARCH
 RUN if [ "$TARGETARCH" = "arm64" ]; then \
-        EXTRA_APT_PACKAGES="erlang ${EXTRA_APT_PACKAGES}"; \
+        EXTRA_APT_PACKAGES="erlang libhdf5-serial-dev pkg-config ${EXTRA_APT_PACKAGES}"; \
     fi;\
     apt-get update --yes && \
     apt-get install --yes --no-install-recommends ${EXTRA_APT_PACKAGES} && \
-    apt-get clean && rm -rf /var/lib/apt/lists/* && \
-    mamba install --yes h5py==3.11.0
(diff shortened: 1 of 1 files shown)


mikibonacci:
@unkcpz , @AndresOrtegaGuerrero

danielhollas:
Can I ask, why is this only needed for arm64?

danielhollas:
If the wheel is not available than pip will attempt to install from source distribution (sdist), which presumably involves some compilation. What is the exact error that you're seeing? Perhaps instead of installing this package to the base image, we should instead install the missing dependencies? Since those might be useful to other packages in the future.

For example, with @AndresOrtegaGuerrero we've been discussing adding cmake recently.
https://aiida.discourse.group/t/installing-cmake-in-aiidalab-container/387

danielhollas:
I think installing hdf5 via apt (or conda for that matter) is preferable since then you can choose which h5py version you install when the plugin is installed.

mikibonacci:
I completely agree. I'll modify my PR in respect of this

mikibonacci:
I added the installation of two packages: `libhdf5-serial-dev` and `pkg-config`. I tried to install just the first package and set up the HDF5_DIR env var, but it was not sufficient to allow the afterwards pip installation of h5py. 

Now h5py will be installed without issues when the aiidalab-qe-vibroscopy plugin is installed.

danielhollas:
Thanks @mikibonacci, can you please push your branch into origin repo (`aiidalab/aiidalab-docker-stack/`) and open a new PR from it? This will allow the build and tests to run (they currently don't run from forks). LMK in case you don't have write access to this repo (but I believe you should).

mikibonacci:
Hi @danielhollas, I tried and I have no permission to write in the origin repo, but as soon as you give me permission I will push! Thanks a lot for the help

danielhollas:
@mikibonacci can you try now?

@unkcpz FYI I've added write permissions to the `internal-maintainers` team LMK if that is not desired

image

unkcpz:
> @unkcpz FYI I've added write permissions to the internal-maintainers team LMK if that is not desired

The `internal-maintainers` team was create to do such things (for qeapp originally), it is desired. Thanks for adding it here.

mikibonacci:
@danielhollas , it works now, I just created the PR. thanks!

mikibonacci:
replaced by https://github.com/aiidalab/aiidalab-docker-stack/pull/468

PR #455, opened 2024-05-14
Update base image to python-3.10.11

danielhollas:

PR #468, opened 2024-06-11
Adding in the `base` Dockerfile the installation of hdf5 library and pkg-config 

mikibonacci:
These pkgs allow to then install h5py via pip without issues.

The installation of pkg-config is required, it was not sufficient to just install the hdf5 and then define the HDF5_DIR environment variable.

### DIFF
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -17,7 +17,7 @@ ENV EXTRA_APT_PACKAGES "curl povray rsync build-essential"
 # but we install it here so that we don't have to invoke apt multiple times.
 ARG TARGETARCH
 RUN if [ "$TARGETARCH" = "arm64" ]; then \
-        EXTRA_APT_PACKAGES="erlang ${EXTRA_APT_PACKAGES}"; \
+        EXTRA_APT_PACKAGES="erlang libhdf5-serial-dev pkg-config ${EXTRA_APT_PACKAGES}"; \
     fi;\
     apt-get update --yes && \
     apt-get install --yes --no-install-recommends ${EXTRA_APT_PACKAGES} && \


unkcpz:
I did discuss with @mikibonacci in person last Friday. I totally agree this is a better solution to add the more generic packages. Just take a look at the failed build on arm64. It is a bit weird to find in the logs around the `mamba install` command there are:

```
#67 51.75    - aiida-core==2.5.1
#67 51.75    - aiida-core=2.5.1
```

Also worth to notice quite a lot conda packages are forced to updated, is this all expected? @danielhollas

danielhollas:
>  Just take a look at the failed build on arm64. 

The build timed-out after two hours. I restarted it now and it went fine so I think this was just Github Actions being slow again, I needed to restart the build on another PR today as well.

```
#67 51.75    - aiida-core==2.5.1
#67 51.75    - aiida-core=2.5.1
```

I think this is because we specify the aiida-core version both via `mamba install` parameter explicitly and also via the constraint file. I can tweak it in another PR to get rid of these confusing lines.

> Also worth to notice quite a lot conda packages are forced to updated, is this all expected?

I am not sure but I am not too surprised since we're installing a relatively recent aiida-core version into an old jupyter-stack image. I am hoping that once we update the image to a more recent one with Python 3.10 (working on that in #455) there will be fewer packages that need upgrading.

danielhollas:
>  It is a bit weird to find in the logs around the mamba install command there are

I fixed this in #461 as part of this commit https://github.com/aiidalab/aiidalab-docker-stack/pull/461/commits/4c0c3b509329e62798c4e5dc025923e91d125afd

danielhollas:
@mikibonacci I am planning to release a new version of the image that will include your change soon, hopefully this week. It would be great if you could pull `docker.io/aiidalab/full-stack:edge` image and test that you can install QeApp and vibroscopy plugin. It would be even more awesome if you could also test running some short calculation to verify nothing important broke, since I've been doing quite a few changes since the last release. Thank you!

mikibonacci:
@danielhollas, I will do it today (is it already pushed?) and let you know if something breaks

danielhollas:
Thanks! The `edge` tag always points to the last commit on the main branch in this repo.

mikibonacci:
yeah, I just saw it after my comment! I did a test for the aiidalab-qe-vibroscopy plugin and it was fine, so I think everything is then working properly for the `edge` tag

PR #469, opened 2024-06-11
Pin jupyter-notebook version

danielhollas:
In https://github.com/aiidalab/aiidalab-qe/issues/640, @superstar54 and I were debugging an issue with a stalled pip installation, likely caused by his quite complex dev environment. One thing I noticed that while the pip was trying hard to solve the dependencies, it started backtracking and looking at older versions of some of the core Jupyter libraries, most notably Jupyter notebook (`notebook` package). This is definitely not desirable, and if pip was able to finish, it would likely break the container. 

We've been having discussions in the past about pinning some jupyter-related packages to prevent this scenario from happening. I'll open a separate issue for that, but for starters I think we should pin at least the notebook package. We definitely don't want users to be able to downgrade the version, and we also don't want them to upgrade it, since we know that v7 will currently not work as it is a completely different architecture.

### DIFF
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -26,29 +26,34 @@ RUN if [ "$TARGETARCH" = "arm64" ]; then \
 WORKDIR /opt/
 
 ARG AIIDA_VERSION
-# Pin shared requirements in the base environment.
-# We pin aiida-core to the exact installed version,
-# to prevent accidental upgrade or downgrade, that might
+
+# Pin certain shared requirements in the base environment
+# so that user cannot change their version and accidently break themselves.
+# Pin aiida-core to the exact installed version
+# to prevent accidental upgrade or downgrade that might
 # induce DB migration or break shared dependencies of AiiDAlab Apps.
 RUN echo "aiida-core==${AIIDA_VERSION}" > /opt/requirements.txt
 
-# Install aiida-core and other shared requirements.
-RUN mamba install --yes \
-     mamba-bash-completion \
-     --file /opt/requirements.txt \
-     && mamba clean --all -f -y && \
-     fix-permissions "${CONDA_DIR}" && \
-     fix-permissions "/home/${NB_USER}"
+# Pin jupyter-notebook to prevent downgrades or upgrades to v7
+RUN echo "notebook==$(jupyter-notebook --version)" >> /opt/requirements.txt
 
-# Pin shared requirements in the conda base environment.
 RUN cat /opt/requirements.txt | xargs -I{} conda config --system --add pinned_packages {}
 
-# Upgrade pip to latest
-RUN pip install --upgrade --no-cache-dir pip
-# Configure pip to use requirements file as constraints file.
+# Configure pip to use the same requirements file as constraints file.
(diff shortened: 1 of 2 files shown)


stack/base/Dockerfile
```diff
@@ -26,30 +26,35 @@ RUN if [ "$TARGETARCH" = "arm64" ]; then \
...
-     mamba-bash-completion \
-     --file /opt/requirements.txt \
-     && mamba clean --all -f -y && \
-     fix-permissions "${CONDA_DIR}" && \
-     fix-permissions "/home/${NB_USER}"
+# Pin jupyter-notebook to prevent downgrades or upgrades to v7
+RUN echo "notebook==$(jupyter-notebook --version)" >> /opt/requirements.txt
 
-# Pin shared requirements in the conda base environment.
 RUN cat /opt/requirements.txt | xargs -I{} conda config --system --add pinned_packages {}
```
danielhollas, line 40: I've rearranged things a bit so that we define the pinned packages first (both for pip and conda) and only then we do any installation.

stack/base/Dockerfile
```diff
@@ -26,30 +26,35 @@ RUN if [ "$TARGETARCH" = "arm64" ]; then \
...
 
-# Upgrade pip to latest
-RUN pip install --upgrade --no-cache-dir pip
```
danielhollas, line 47: pip is originally installed via conda so I've changed this and use `mamba update pip` instead.

danielhollas: Perhaps we should think about updating conda and mamba as well?

danielhollas:
@unkcpz would you have time to take a look in the next few days? I'd like to publish a new version of the image after we merge this.

unkcpz:
Yes, it is in my list, I'll have a look asap. Sorry.

danielhollas:
No worries at all, thanks!

stack/base/Dockerfile
unkcpz: Make sense!

stack/base/Dockerfile
unkcpz: The mamba is from jupyter image, so I think it is safe to use any version it used, and if I remember correctly they spin up the build process periodically and will try to bring the latest mamba.

unkcpz (approved):
Looks like a safe move. Thanks @danielhollas.

danielhollas:
Thanks @unkcpz. I've added some tests. After merging this I'm going to publish a new version.

stack/base/Dockerfile
unkcpz: I didn't use aiidalab mamba for quite a while, did you see some annoying warnings ask to update? If so I am totally fine with updating to the latest.

unkcpz (approved):

stack/base/Dockerfile
danielhollas: I did not, I just think it is a good practice to keep it updated. I'll take a closer look and prepare a PR at some point.

btw: Thanks for the review, I've added some tests in case you want to take a look. Waiting for the build to finish and will then merge.

Issue #407, opened 2023-08-31
Set caching as default for the profile config


unkcpz:
We find caching is quite useful, but potentially may have some problems. But we want to make it default and have a widget to control the switch on/off it.

danielhollas:
Thanks for creating an issue.

I wonder if we should have default OFF first for some initial testing?
Also I believe that AiiDA 2.4 had some important caching fixes so #396 needs to be merged first.

danielhollas:
I think we should do this after 2.6 is released, which contains another set of caching improvements / fixes. Hopefully things should be more stable from that point.

danielhollas:
@edan-bainglass @yakutovicha we should consider enabling caching by default in the new image. I've been using it in my own app deployment for quite a while and haven't noticed any problems.

edan-bainglass:
> [@edan-bainglass](https://github.com/edan-bainglass) [@yakutovicha](https://github.com/yakutovicha) we should consider enabling caching by default in the new image. I've been using it in my own app deployment for quite a while and haven't noticed any problems.

Thanks @danielhollas. I'll have a chat with Julian and Ali in the AiiDA team regarding caching, as they both put in some work there last year. But indeed, it is a powerful AiiDA feature that is lost on those who are unaware. I support enabling as default barring any critical comments from the AiiDA team.

PR #474, opened 2024-07-02
Update aiida-core to 2.6.2

danielhollas:
Looks like we're good to go here. I've launched tests for AWB in https://github.com/aiidalab/aiidalab-widgets-base/pull/616 and for QeApp in https://github.com/aiidalab/aiidalab-qe/pull/760 and they all passed.

Once we update, I think we can enable caching by default, per #407.

### DIFF
--- a/build.json
+++ b/build.json
@@ -10,7 +10,7 @@
       "default": "3.9.13"
     },
     "AIIDA_VERSION": {
-      "default": "2.5.1"
+      "default": "2.6.2"
     },
     "AIIDALAB_VERSION": {
       "default": "24.07.0"
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -10,8 +10,11 @@ def test_notebook_service_available(notebook_service):
 
 def test_verdi_status(aiidalab_exec, nb_user):
     output = aiidalab_exec("verdi status", user=nb_user).strip()
-    assert "Connected to RabbitMQ" in output
+    for status in ("version", "config", "profile", "storage", "broker", "daemon"):
+        assert f"✔ {status}" in output
+    assert "/home/jovyan/.aiida" in output
     assert "Daemon is running" in output
+    assert "Unable to connect to broker" not in output
 
 
 def test_ssh_agent_is_running(aiidalab_exec, nb_user):


unkcpz:
I think @sphuber will firing another patch release? Let's wait a bit for that. 

> I think we can enable caching by default

Let's hold a bit on this. For PH calculation and bands calculation, we are definitely not ready to move forward on this. But I think we need to start testing on it more, to make the problem more exposed. If you are interested, please have a look at the AEP (https://github.com/aiidateam/AEP/pull/35) and the initial discussion (https://github.com/aiidateam/aiida-core/issues/5178).

I had some workarounds with changing the workchain logic to catch the remote data clean up issue to rerun the parent calcualtion, but far from perfect. I'll update the AEP and make a slide to try to describe the problem more clear.

danielhollas:
> I think sphuber will firing another patch release? Let's wait a bit for that.

Aiida 2.6.2 has been released today. :tada: This should be ready to go.

The new `latest` image release needs to wait until https://github.com/aiidalab/aiidalab/issues/442 is fixed, I'm hoping to get that done this week so I can deploy the new image on our server.

@unkcpz @superstar54 LMK if you're generally available for reviews or if I should be reaching out to somebody else in the meantime.

unkcpz (approved):
Thanks @danielhollas, I think it is ready to go.

edan-bainglass:
@danielhollas @unkcpz I'm unable to install aiida-core>2.5.1 on aiidalab-fullstack containers.

```
(base) (master) aiidalab-widgets-base > pip install aiida-core~=2.6
ERROR: Cannot install aiida-core~=2.6 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested aiida-core~=2.6
    The user requested (constraint) aiida-core==2.5.1

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

unkcpz:
Since we had pinned the `aiida-core` version in the container. If you gonna to use `2.6.1` you need to update the image.

edan-bainglass:
I suspected as much and pulled the new version. Same issue 😭

unkcpz:
Please try pull `full-stack:edge`, we not yet release the image.

PR #475, opened 2024-07-03
Updating the *before_notebook* `20_start-postgresql.sh` and `40_prepare_aiida.sh`

mikibonacci:
Fixings are done as for arm64 I encountered issues in restarting the container.

1 - Deleting old ${PSQL_LOGFILE}.1.gz file before trying to compressed again in `20_start-postgresql.sh`. Otherwise, this can give errors and so the container will exit with non-zero status. 
2 - Clean stop of the daemon before the "verdi storage migrate --force" in `40_prepare_aiida.sh`. This is done because sometimes the restart of the docker container will give error in trying to migrate, as the daemon is still running. I am not able to reproduce the error in a standard way, however.


mikibonacci:
I forgot that the test are not working if the PR is from outside. I fill create the PR from this repository.

PR #476, opened 2024-07-03
Updating the *before_notebook* `20_start-postgresql.sh` and `40_prepare_aiida.sh`

mikibonacci:
Fixings are done as for arm64 I encountered issues in restarting the container.

1 - Deleting old ${PSQL_LOGFILE}.1.gz file before trying to compressed again in `20_start-postgresql.sh`. Otherwise, this can give errors and so the container will exit with non-zero status. 
2 - Clean stop of the daemon before the "verdi storage migrate --force" in `40_prepare_aiida.sh`. This is done because sometimes the restart of the docker container will give error in trying to migrate, as the daemon is still running. I am not able to reproduce the error in a standard way, however.

### DIFF
--- a/stack/base-with-services/before-notebook.d/20_start-postgresql.sh
+++ b/stack/base-with-services/before-notebook.d/20_start-postgresql.sh
@@ -22,10 +22,7 @@ else
     chmod -R g-rwxs "${PGDATA}"
 
     if [[ -f ${PGDATA}/logfile ]]; then
-        if [[ -f ${PSQL_LOGFILE}.1.gz ]]; then
-            echo "Deleting old ${PSQL_LOGFILE}.1.gz compressed file"
-            rm "${PSQL_LOGFILE}.1.gz"
-        fi
+        rm -f "${PSQL_LOGFILE}.1.gz" 
         mv "${PSQL_LOGFILE}" "${PSQL_LOGFILE}.1" && gzip "${PSQL_LOGFILE}.1"
     fi
     # Cleaning up the mess if PostgreSQL was not shutdown properly.
--- a/stack/base-with-services/before-notebook.d/20_start-postgresql.sh
+++ b/stack/base-with-services/before-notebook.d/20_start-postgresql.sh
@@ -22,7 +22,11 @@ else
     chmod -R g-rwxs "${PGDATA}"
 
     if [[ -f ${PGDATA}/logfile ]]; then
-         mv "${PSQL_LOGFILE}" "${PSQL_LOGFILE}.1" && gzip "${PSQL_LOGFILE}.1"
+        if [[ -f ${PSQL_LOGFILE}.1.gz ]]; then
+            echo "Deleting old ${PSQL_LOGFILE}.1.gz compressed file"
+            rm "${PSQL_LOGFILE}.1.gz"
+        fi
+        mv "${PSQL_LOGFILE}" "${PSQL_LOGFILE}.1" && gzip "${PSQL_LOGFILE}.1"
     fi
     # Cleaning up the mess if PostgreSQL was not shutdown properly.
     rm -vf "${PGDATA}/postmaster.pid"
--- a/stack/base-with-services/before-notebook.d/20_start-postgresql.sh
+++ b/stack/base-with-services/before-notebook.d/20_start-postgresql.sh
@@ -22,7 +22,7 @@ else
     chmod -R g-rwxs "${PGDATA}"
 
     if [[ -f ${PGDATA}/logfile ]]; then
-        rm -f "${PSQL_LOGFILE}.1.gz" 
+        rm -f "${PSQL_LOGFILE}.1.gz"
         mv "${PSQL_LOGFILE}" "${PSQL_LOGFILE}.1" && gzip "${PSQL_LOGFILE}.1"
     fi
     # Cleaning up the mess if PostgreSQL was not shutdown properly.
--- a/stack/base/before-notebook.d/40_prepare-aiida.sh
+++ b/stack/base/before-notebook.d/40_prepare-aiida.sh
@@ -76,7 +76,7 @@ else
   # Migration will run for the default profile.
   pgrep -af 'verdi.* daemon' && echo "ERROR: AiiDA daemon is already running!" && exit 1
   ## clean up stale PID-files -> .aiida/daemon/circus-{profile-name}.pid
-  rm -f "/home/${NB_USER}/.aiida/daemon/circus-*.pid"
+  rm -f /home/${NB_USER}/.aiida/daemon/circus-*.pid
   verdi storage migrate --force
 
 fi
--- a/stack/base/before-notebook.d/40_prepare-aiida.sh
+++ b/stack/base/before-notebook.d/40_prepare-aiida.sh
@@ -74,8 +74,8 @@ load_computer('${computer_name}').set_minimum_job_poll_interval(${job_poll_inter
 else
 
   # Migration will run for the default profile.
-  ## We need to stop the daemon before.
-  verdi daemon stop
+  ## clean up stale PID-files -> .aiida/daemon/circus-{profile-name}.pid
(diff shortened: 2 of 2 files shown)


stack/base-with-services/before-notebook.d/20_start-postgresql.sh
```diff
@@ -22,7 +22,11 @@ else
     chmod -R g-rwxs "${PGDATA}"
 
     if [[ -f ${PGDATA}/logfile ]]; then
-         mv "${PSQL_LOGFILE}" "${PSQL_LOGFILE}.1" && gzip "${PSQL_LOGFILE}.1"
+        if [[ -f ${PSQL_LOGFILE}.1.gz ]]; then
+            echo "Deleting old ${PSQL_LOGFILE}.1.gz compressed file"
+            rm "${PSQL_LOGFILE}.1.gz"
+        fi
+        mv "${PSQL_LOGFILE}" "${PSQL_LOGFILE}.1" && gzip "${PSQL_LOGFILE}.1"
```
danielhollas, line 29:
```suggestion
        rm -f "${PSQL_LOGFILE}.1.gz" 
        mv "${PSQL_LOGFILE}" "${PSQL_LOGFILE}.1" && gzip "${PSQL_LOGFILE}.1"
```

stack/base/before-notebook.d/40_prepare-aiida.sh
```diff
@@ -74,6 +74,8 @@ load_computer('${computer_name}').set_minimum_job_poll_interval(${job_poll_inter
 else
 
   # Migration will run for the default profile.
+  ## We need to stop the daemon before.
+  verdi daemon stop
```
danielhollas, line 78: I think we need to understand why this is needed, since the daemon cannot be running at the container startup.

@sphuber does `verdi daemon stop` do some kind of cleanup that might explain why it is needed here? It is true that we currently don't gracefully stop the daemon when we stop the container so there might perhaps be some stale pid files or something?

@mikibonacci what is the exact error that you saw coming from the `verdi storage migrate` command?

danielhollas (changes requested):

danielhollas:
Thanks for catching these issues.

>  I encountered issues in restarting the container.

How exactly are you restarting the container?

stack/base/before-notebook.d/40_prepare-aiida.sh
sphuber: Depends on the version. The behavior has changed a bit in the 2.x minor releases. But currently, yes, `verdi daemon stop` should clean up stale PID-files. And it is possible that some other commands will check if the daemon "is running" based on that pid file. Would have to look into more detail if you know exactly which version you are targeting

danielhollas: Thanks! We're currently on 2.5, and hopefully will switch to 2.6 soon.

I was trying to get rid of `verdi daemon stop` because it adds 2s to container startup. Perhaps we can just remove the stale pid files manually?

mikibonacci: Hi @danielhollas  and @sphuber, the exact error printed in the log is:

`2024-07-03 15:41:58 Critical: Migration aborted, the daemon for the profile is still running.`

The target AiiDA version is `aiida-core==2.5.1` (build=pyhca7485f_0, channel=conda-forge) indeed.

mikibonacci:
For the restart, I used `aiidalab-launch` and also the manual startup (docker desktop and terminal). These issues are there only when I use arm64 machine (my Mac). That's strange.

stack/base/before-notebook.d/40_prepare-aiida.sh
sphuber: > I was trying to get rid of verdi daemon stop because it adds 2s to container startup. Perhaps we can just remove the stale pid files manually?

Well if you are absolutely certain that the daemon shouldn't be running, you could just manually remove the pid file `.aiida/daemon/circus-{profile-name}.pid` which should be very fast. The `verdi daemon stop` command will actually try to reach the daemon, which has a 2 second timeout by default.

stack/base/before-notebook.d/40_prepare-aiida.sh
```diff
@@ -74,6 +74,8 @@ load_computer('${computer_name}').set_minimum_job_poll_interval(${job_poll_inter
 else
 
   # Migration will run for the default profile.
+  ## clean up stale PID-files -> .aiida/daemon/circus-{profile-name}.pid
+  rm -f "/home/${NB_USER}/.aiida/daemon/circus-${AIIDA_PROFILE_NAME}.pid"
```
mikibonacci, line 78: @danielhollas we suppose the user will just use the ${AIIDA_PROFILE_NAME} profile, right? 
I am not sure if we should also consider the possibility to define other profiles, so we should consider also to clean their pid(s)

stack/base/before-notebook.d/40_prepare-aiida.sh
unkcpz: Command `verdi storage migrate` has access to daemon client, so probably another solution is when `--force` flag is used, it does clean stale PID files before migrate.

stack/base/before-notebook.d/40_prepare-aiida.sh
danielhollas: Let's just do:

```suggestion
  rm -f "/home/${NB_USER}/.aiida/daemon/circus-*.pid"
```

stack/base/before-notebook.d/40_prepare-aiida.sh
sphuber: I am not sure that it would be a good idea to have --force ignore a running daemon. The --force flag is there to not have to type out the confirmation message. This has been the case for a long time and users may have grown used to that. If we now also have it start ignoring a running daemon, that might be dangerous.

stack/base/before-notebook.d/40_prepare-aiida.sh
```diff
@@ -74,6 +74,8 @@ load_computer('${computer_name}').set_minimum_job_poll_interval(${job_poll_inter
 else
 
   # Migration will run for the default profile.
```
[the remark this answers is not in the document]
danielhollas: Yeah, I think less-specific here is more "safe", it's essentially an assert the no verdi daemon process is running. Thanks!

danielhollas:
@mikibonacci can you test the latest version of the image (`ghcr.io/aiidalab/full-stack:pr-476`).

Once you confirm that it works reliably for you I'll merge and release a new version.

Thanks for catching this and the fixes!

stack/base/before-notebook.d/40_prepare-aiida.sh
unkcpz: Okay, then manually delete it as Miki did here is fine.

stack/base/before-notebook.d/40_prepare-aiida.sh
```diff
@@ -74,6 +74,9 @@ load_computer('${computer_name}').set_minimum_job_poll_interval(${job_poll_inter
...
+  pgrep -af 'verdi.* daemon' && echo "ERROR: AiiDA daemon is already running!" && exit 1
+  ## clean up stale PID-files -> .aiida/daemon/circus-{profile-name}.pid
+  rm -f "/home/${NB_USER}/.aiida/daemon/circus-*.pid"
```
danielhollas, line 79: Hmm, looks like this command doesn't actually remove the pid file? Let's try without the quotes.

```suggestion
  rm -f /home/${NB_USER}/.aiida/daemon/circus-*.pid
```

mikibonacci:
Hi @danielhollas , I tested the `ghcr.io/aiidalab/full-stack:pr-476` and it works well for me! Thanks a lot for the review!

danielhollas (approved):

danielhollas:
Thanks for testing @mikibonacci. I'll release a new version soon with your fix.

I've opened #477 as a follow up to this to prevent these issues in the future.

danielhollas:
@mikibonacci new version is released. :rocket:

PR #479, opened 2024-07-14
Switch to nbclassic and Python 3.12.11

danielhollas:
fixes #552 

We switch to nbclassic interface, which is now supported in appmode.

Notebook v7 is currently not supported in Appmode.


edan-bainglass (commented):
Left comments

stack/base/Dockerfile
```diff
@@ -60,7 +60,7 @@ COPY pip.conf "${CONDA_DIR}/pip.conf"
 # Update async_generator, certipy to satisfy `pip check`
 # https://github.com/aiidalab/aiidalab-docker-stack/issues/490
 # Install aiida-core and other shared requirements.
-RUN mamba update -y pip async_generator certipy && \
+RUN mamba update -y pip zstandard async_generator certipy && \
```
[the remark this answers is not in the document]
yakutovicha: @danielhollas, what was the reason for this change?

stack/lab/Dockerfile
```diff
@@ -32,8 +31,8 @@ RUN echo 'eval "$(_AIIDALAB_COMPLETE=bash_source aiidalab)"' >> "${CONDA_DIR}/et
 ARG AIIDALAB_HOME_VERSION
 RUN git clone https://github.com/aiidalab/aiidalab-home && \
      cd aiidalab-home && \
-     git checkout v"${AIIDALAB_HOME_VERSION}" && \
-     pip install --no-user --quiet --no-cache-dir "./" && \
+     git checkout "${AIIDALAB_HOME_VERSION}" && \
+     pip install --no-user --no-cache-dir "./" && \
```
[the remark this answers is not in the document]
yakutovicha: I used it for debugging, just to have more info in the output when installing the home app. Can revert.

yakutovicha:
I moved the parallel build of arm/amd architectures to another PR: https://github.com/aiidalab/aiidalab-docker-stack/pull/553, so we don't do everything here.

stack/base-with-services/before-notebook.d/20_start-postgresql.sh
```diff
@@ -10,12 +10,14 @@ PSQL_LOGFILE="${PGDATA}/logfile"
 # -w waits until server is up
 PSQL_START_CMD="pg_ctl --timeout=180 -w -l ${PSQL_LOGFILE} start"
 
-MAMBA_RUN="mamba run -n aiida-core-services"
+# Activate the aiida-core-services environment directly instead of using mamba run
```
danielhollas, line 13: I am very confused, do we know why `mamba run` suddenly doesn't work?

stack/base-with-services/before-notebook.d/30_start-rabbitmq.sh
```diff
@@ -4,7 +4,14 @@ set -emx
...
+conda activate aiida-core-services
+
 # NOTE: In arm64 build, rabbitmq is not installed via conda,
 # but the following incantation still works since
 # rabbitmq-server is available globally.
-mamba run -n aiida-core-services rabbitmq-server -detached
+rabbitmq-server -detached
+
+# Wait a moment for RabbitMQ to start
+sleep 2
```
danielhollas: Is this really needed or was this for debugging? We spent a lot of time trying the make the container startup as fast as possible so would like to avoid sleeps unless really necessary.

stack/docker-compose.base-with-services.yml
```diff
@@ -1,17 +1,21 @@
...
-version: '3.4'
-
 services:
 
     aiidalab:
         image: ${REGISTRY:-}${BASE_WITH_SERVICES_IMAGE:-aiidalab/base-with-services}${VERSION:-}
         environment:
             TZ: Europe/Zurich
-            DOCKER_STACKS_JUPYTER_CMD: notebook
+            DOCKER_STACKS_JUPYTER_CMD: nbclassic
```
danielhollas: Let's remove this, I don't see a reason why this should be defined here, it's already defined in Dockerfile and should not really be changed.
```suggestion
```

stack/docker-compose.base-with-services.yml
```diff
@@ -1,17 +1,21 @@
...
             SETUP_DEFAULT_AIIDA_PROFILE: 'true'
             AIIDALAB_DEFAULT_APPS: ''
         volumes:
             - aiidalab-home-folder:/home/jovyan
+        healthcheck:
```
danielhollas, line 11: Can you explaing the rationale for adding this?

tests/test_base.py
```diff
@@ -8,10 +8,11 @@
 
 
 @pytest.fixture
-def venv(tmp_path, aiidalab_exec):
-    venv_path = tmp_path / ".venv"
-    aiidalab_exec(f"python -m venv {venv_path}")
-    return venv_path
+def venv(aiidalab_exec, nb_user):
```
danielhollas: Why this change? Previous code seems preferable and simpler, `tmp_path` fixture already takes care of the cleanup

danielhollas (commented):
Some comments / questions

build.json
danielhollas:
```suggestion
      "default": "main"
```

danielhollas:
I've merged in main to get aiida-core 2.8.0

stack/docker-compose.base-with-services.yml
danielhollas: Done in #556

stack/base-with-services/before-notebook.d/20_start-postgresql.sh
yakutovicha: I don't know why, but my observation is that `mamba run` hangs when it calls a process that launches another process in the background. For some reason, it also waits for the background process to complete, which never happens. So the script never proceeds to completion.

~This, however, only happened for RQM; the change in PostgreSQL was only done so that RMA and PSQL launchers look similar~. [EDITED]

stack/base/Dockerfile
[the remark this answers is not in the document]
yakutovicha: not needed, in the end. So removed here: https://github.com/aiidalab/aiidalab-docker-stack/pull/479/commits/0fe29f209f943446d30e250b3f34ff5c6468a2be.

yakutovicha:
For the record, the failing tests are because we are constraining `ipywidgets~=8.0`, which means we won't be able to make the tests run before apps (QE app in this case) migrate to ipywidgets 8.x.

stack/lab/Dockerfile
```diff
@@ -80,7 +77,8 @@ USER ${NB_USER}
 
 WORKDIR "/home/${NB_USER}"
 
-RUN  mkdir -p /home/${NB_USER}/apps
+RUN mkdir -p /home/${NB_USER}/apps
+RUN mkdir -p /home/${NB_USER}/.tmp_notebooks
```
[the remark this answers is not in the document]
yakutovicha: no:

Image

I would then suggest removing that change and trying without it. If needed later, we can do a follow-up PR to not overload this one.

stack/lab/Dockerfile
```diff
@@ -44,12 +43,10 @@ RUN ipywidgets_major=$(mamba list -f ipywidgets | tail -1 | awk '{print $2}' | c
      conda config --system --add pinned_packages "ipywidgets=${ipywidgets_major}"
 
 # Install and enable appmode, turning Jupyter notebooks to Apps
-RUN pip install appmode==0.8.0 --no-cache-dir --no-user
-# Enable appmode extension
-RUN jupyter nbextension enable --py --sys-prefix appmode && \
-     jupyter serverextension enable --py --sys-prefix appmode
+RUN pip install appmode==1.1.0 --no-cache-dir --no-user
```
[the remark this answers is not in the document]
yakutovicha: Indeed, seems not needed anymore. Removed in https://github.com/aiidalab/aiidalab-docker-stack/pull/479/commits/060ed4a248780b1df808668694965aa556283aca.

stack/lab/Dockerfile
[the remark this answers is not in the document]
yakutovicha: I checked the temporary notebooks created by appmode - they still live in the same folder as the source notebook. The name changed slightly, but that change has no consequence for us.

All in all, I see no need for a separate folder, as things are handled pretty much the same way as before.

I remoted the folder creation here: https://github.com/aiidalab/aiidalab-docker-stack/pull/479/commits/8a51479bf7ac18d5303853a076f9da978ea69e66

tests/test_base.py
yakutovicha: Reverted in https://github.com/aiidalab/aiidalab-docker-stack/pull/479/commits/bd7992a9ba44985f8f05e797fe39deb408c7fa9f.

stack/docker-compose.base-with-services.yml
yakutovicha: Yes. Before this was added (https://github.com/aiidalab/aiidalab-docker-stack/pull/479/commits/1a6fb5c86e0bc814238836e8688924b73fb3148c), all tests were failing because the standard health check was too restrictive: on GitHub runners, there was not enough time for everything to get started. As a result, the container was marked as unhealthy.

For example, all tests from the [previous commit](https://github.com/aiidalab/aiidalab-docker-stack/actions/runs/23460052563/job/68259365021) failed exactly for that reason.

stack/base-with-services/before-notebook.d/30_start-rabbitmq.sh
yakutovicha: I think you are right - it is not needed. None of the following startup scripts requires reassurance that RMQ is running. Deleted in https://github.com/aiidalab/aiidalab-docker-stack/pull/479/commits/d9609f9eb4ae5a9ec3541368881adc93f31f2d1a.

yakutovicha:
@edan-bainglass, @danielhollas: Here is the result of the investigation regarding `mamba run` vs `conda activate`.

By default, conda run sets up pipes for stdout/stderr to capture the subprocess output. The flow is:

  1. `mamba run -n aiida-core-services pg_ctl start` spawns `pg_ctl` as a child process
  2. `pg_ctl`'s stdout/stderr are connected to pipes back to mamba run
  3. mamba run reads from those pipes until EOF, then exits

**What happens when starting the deamons**

When `pg_ctl` starts or `rabbitmq-server -detached` runs:

  1. The command forks a long-running daemon process.
  2. The daemon inherits the pipe file descriptors from its parent (stdout/stderr).
  3. The parent process exits normally.
  4. `mamba run` sees the parent exit, but it's still waiting for EOF on the pipe.
  5. EOF on a pipe only occurs when every process holding the write end closes it.
  6. The daemon still holds the pipe's write end open (inherited at fork).
  7. The daemon never exits -> pipe never gets EOF -> mamba run hangs forever

An alternative solution is suggested [here](https://github.com/conda/conda/issues/9412) is to use `--no-capture-output`

Here is the summary of all tries:


Approach | Result | Notes
-- | -- | --
mamba run (original) | HANGS | Pipe inheritance by forked daemons
mamba run --no-capture-output | BROKEN | Generates exec (in wrapper script -> exec): invalid option (mamba bug)
conda run --no-capture-output | WORKS | Avoids pipe capture without the exec bug
conda activate + direct commands | WORKS | To me, this is the cleanest way.

stack/base-with-services/before-notebook.d/20_start-postgresql.sh
yakutovicha: here is the investigation results: https://github.com/aiidalab/aiidalab-docker-stack/pull/479#issuecomment-4205169211

stack/base-with-services/before-notebook.d/20_start-postgresql.sh
```diff
@@ -10,12 +10,12 @@ PSQL_LOGFILE="${PGDATA}/logfile"
 # -w waits until server is up
 PSQL_START_CMD="pg_ctl --timeout=180 -w -l ${PSQL_LOGFILE} start"
 
-MAMBA_RUN="mamba run -n aiida-core-services"
+CONDA_RUN="conda run -n aiida-core-services"
```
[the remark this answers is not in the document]
yakutovicha: Here it is: https://github.com/aiidalab/aiidalab-docker-stack/pull/479#issuecomment-4205169211

yakutovicha:
@danielhollas, @edan-bainglass - to me, the PR is ready to go, please re-review.

stack/base-with-services/before-notebook.d/20_start-postgresql.sh
danielhollas: Thanks for the detailed investigation! Great that you got to the bottom of it. 
I'll have a closer look but regardless of what solution we decide, I'd revert the changes to the start-postgresql script as they are not needed and I prefer the `mamba run` approach in general (running eval is not great and can have other side effects)

danielhollas:
@yakutovicha thanks so much, I'll have a look (but will not have time today). Note that this is still blocked on the RabbitMQ upgrade, we need to resolve #510 first. But I agree that otherwise this looks ready.

Can you please merge main? That should remove some of the changes to the docker compose files.

stack/base-with-services/before-notebook.d/20_start-postgresql.sh
yakutovicha: I re-ran the tests with `mamba run` for PSQL -- the problem is the same (pipe inheritance issue), so I remembered it wrongly. The changes are needed in both startup scripts.

I edited my comment above.

danielhollas:
@yakutovicha I just launched the latest image from this PR and I am seeing a Notebook v7 warning banner:

image

I believe it was hidden in the earlier versions of this PR, perhaps that change got lost?

EDIT: Adding `"--NotebookApp.show_banner=False "\` to `NOTEBOOK_ARGS` should resolve this, see: https://github.com/aiidalab/aiidalab-docker-stack/pull/455/changes

build.json
```diff
@@ -16,7 +16,7 @@
       "default": "26.02.0"
     },
     "AIIDALAB_HOME_TAG": {
-      "default": "v25.02.0"
+      "default": "main"
```
edan-bainglass, line 19: Why unpin?

docker-bake.hcl
```diff
@@ -21,7 +21,7 @@ variable "AIIDALAB_HOME_TAG" {
 }
 
 variable "JUPYTER_BASE_IMAGE" {
-  default = "jupyter/minimal-notebook:python-${PYTHON_VERSION}"
+  default = "quay.io/jupyter/minimal-notebook:python-${PYTHON_VERSION}"
```
edan-bainglass, line 24: Why?

tests/test_lab.py
```diff
@@ -22,7 +22,7 @@ def test_correct_aiidalab_version_installed(package_info, aiidalab_version):
 
 
 def test_correct_aiidalab_home_version_installed(package_info, aiidalab_home_tag):
-    info = package_info("aiidalab-home")
+    info = package_info("aiidalab_home")
```
edan-bainglass, line 25: Was this a typo before?

edan-bainglass (approved):
LGTM. Just a few comments, but good to go.

tests/test_lab.py
danielhollas: Newer mamba version in this image uses this canonicalized package name in its `mamba list` output

build.json
danielhollas: We should make a new release of aiidalab-home

danielhollas:
Pushed some small changes, hope that's okay. Disabled QeApp integration tests for now, but before this is merged it would be good to have at least a rough idea what it would take to upgrade QeApp for this image. 

May I ask, what kind of testing has been done on this image so far? I think at the very least we should test individual AWB widgets.

Marking this as blocked until #510 is merged (hope to make some progress on that over weekend) and we need to first release the 2.8 based image anyway.

danielhollas:
This is very minor, but I am seeing this warning when using pip in this image

image

We can add `ENV PIP_USE_FEATURE="build-constraint"` in the base image to get rid of it (where we set PIP_CONSTRAINT).

docker-bake.hcl
danielhollas: Newer version of jupyter stack image are no longer published on Dockerhub

danielhollas:
> Marking this as blocked until https://github.com/aiidalab/aiidalab-docker-stack/pull/510 is merged (hope to make some progress on that over weekend) and we need to first release the 2.8 based image anyway.

The RMQ update has been merged and I've updated this PR here. The changes here look good to me, but I think we should release a new image with aiida-core 2.8.0 before merging this, see #564. I've also discussed with @yakutovicha that it would be good to do a bit more testing, in particular to have at least one app upgraded and confirmed working, so that we do not run into any further surprises.

Please also address my previous comments: (EDIT: I've done the first two) 
- [x] ~~https://github.com/aiidalab/aiidalab-docker-stack/pull/479#issuecomment-4227314870 (pip deprecation warning)~~ This turned out to be difficult so I've given up for now. This problem should resolve by itself in future pip versions
- [x] https://github.com/aiidalab/aiidalab-docker-stack/pull/479#issuecomment-4214098514 (hiding Notebook 7 banner)
- [ ] https://github.com/aiidalab/aiidalab-docker-stack/pull/479#discussion_r3061893545 (making a new release of aiidalab-home)

stack/docker-compose.base-with-services.yml
danielhollas: I guess the question is then why was this not needed before? Is the image starting slower now? (or is there something different?). Maybe we could try removing this now that the PR is basically done to see if this is still needed?

danielhollas (commented):
I think once we make the `aiidalab-home` release we can merge this.

yakutovicha (approved):

PR #482, opened 2024-07-19
Update RMQ to 3.10, install from conda on arm64

danielhollas:
After a fair amount of work, we have finally arrived: rabbitmq arm64 build is now available on conda-forge! This let's us simplify the build considerably.

### DIFF
--- a/build.json
+++ b/build.json
@@ -7,7 +7,7 @@
       "default": "15"
     },
     "RMQ_VERSION": {
-      "default": "3.9.13"
+      "default": "3.10.25"
     },
     "AIIDA_VERSION": {
       "default": "2.8.0"
--- a/stack/base-with-services/Dockerfile
+++ b/stack/base-with-services/Dockerfile
@@ -8,7 +8,6 @@ WORKDIR /opt/
 
 ARG PGSQL_VERSION
 ARG RMQ_VERSION
-ARG TARGETARCH
 
 # Location of the Postgresql DB
 # This variable is automatically picked up by initdb and pg_ctl
@@ -16,33 +15,11 @@ ARG TARGETARCH
 ENV PGDATA=/home/${NB_USER}/.postgresql
 
 # Install RabbitMQ and PostgreSQL in a dedicated conda environment.
-#
-# RabbitMQ is currently not available on conda-forge for arm64, see:
-# https://github.com/conda-forge/rabbitmq-server-feedstock/issues/67If
-# Instead we need install erlang via apt and RabbitMQ as a "Generic Unix Build", see:
-# https://www.rabbitmq.com/install-generic-unix.html
-# Note that this version must be compatible with system's erlang version.
-# Currently installed Erlang version is 23.3, so the latest supported RMQ version is 3.9.21
-# https://www.rabbitmq.com/docs/which-erlang#old-timers
-# Note that system erlang from arm64 is already installed in the base image,
-# together with other APT dependencies to save build time.
-RUN if [ "$TARGETARCH" = "amd64" ]; then \
-     mamba create -p /opt/conda/envs/aiida-core-services --yes \
-         postgresql=${PGSQL_VERSION} \
-         rabbitmq-server=${RMQ_VERSION} && \
-     mamba clean --all -f -y && \
-     fix-permissions "${CONDA_DIR}"; \
-  elif [ "$TARGETARCH" = "arm64" ]; then \
-     mamba create -p /opt/conda/envs/aiida-core-services --yes \
-         postgresql=${PGSQL_VERSION} && \
-     mamba clean --all -f -y && \
-     wget -c https://github.com/rabbitmq/rabbitmq-server/releases/download/v${RMQ_VERSION}/rabbitmq-server-generic-unix-${RMQ_VERSION}.tar.xz && \
-     tar -xf rabbitmq-server-generic-unix-${RMQ_VERSION}.tar.xz && \
-     rm rabbitmq-server-generic-unix-${RMQ_VERSION}.tar.xz && \
-     mv rabbitmq_server-${RMQ_VERSION} /opt/conda/envs/aiida-core-services/rabbitmq_server && \
-     ln -sf /opt/conda/envs/aiida-core-services/rabbitmq_server/sbin/* /opt/conda/envs/aiida-core-services/bin/ && \
-     fix-permissions "${CONDA_DIR}"; \
-  fi
+RUN mamba create -p /opt/conda/envs/aiida-core-services --yes \
+        postgresql=${PGSQL_VERSION} \
+        rabbitmq-server=${RMQ_VERSION} && \
+    mamba clean --all -f -y && \
+    fix-permissions "${CONDA_DIR}"
 
 # Configure AiiDA profile.
 COPY config-quick-setup.yaml .
--- a/stack/base-with-services/before-notebook.d/30_start-rabbitmq.sh
+++ b/stack/base-with-services/before-notebook.d/30_start-rabbitmq.sh
@@ -4,7 +4,4 @@ set -emx
 # Fix issue where the erlang cookie permissions are corrupted.
 chmod 400 "/home/${NB_USER}/.erlang.cookie" || echo "erlang cookie not created yet."
 
-# NOTE: In arm64 build, rabbitmq is not installed via conda,
-# but the following incantation still works since
-# rabbitmq-server is available globally.
 mamba run -n aiida-core-services rabbitmq-server -detached
--- a/stack/base-with-services/configure_rabbitmq.sh
+++ b/stack/base-with-services/configure_rabbitmq.sh
@@ -1,14 +1,9 @@
 #!/bin/bash
 set -emx
 
-RMQ_ETC_DIR_ARM64="/opt/conda/envs/aiida-core-services/rabbitmq_server/etc/rabbitmq"
-RMQ_ETC_DIR_AMD64="/opt/conda/envs/aiida-core-services/etc/rabbitmq"
-if [[ -d $RMQ_ETC_DIR_ARM64 ]]; then
-    RMQ_ETC_DIR="$RMQ_ETC_DIR_ARM64"
-elif [[ -d $RMQ_ETC_DIR_AMD64 ]]; then
-    RMQ_ETC_DIR="$RMQ_ETC_DIR_AMD64"
-else
-    echo "ERROR: Could not find RabbitMQ etc directory"
+RMQ_ETC_DIR="/opt/conda/envs/aiida-core-services/etc/rabbitmq"
+if [[ ! -d $RMQ_ETC_DIR ]]; then
+    echo "ERROR: directory '$RMQ_ETC_DIR' does not exist!"
     exit 1
 fi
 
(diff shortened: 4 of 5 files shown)


stack/lab/before-notebook.d/60_prepare-aiidalab.sh
```diff
@@ -6,11 +6,6 @@ set -x
 # Environment.
 export SHELL=/bin/bash
 
-# Fix https://github.com/aiidalab/aiidalab-docker-stack/issues/225
```
danielhollas, line 9: This issue has been fixed a long time ago and should not be relevant in the new docker stack.

danielhollas:
@mikibonacci could you please test the full-stack image from this PR on your Mac Arm machine? I am specifically interested to see if rabbitmq works correctly, so if you could install the QeApp and run some quantum espresso calculations that would be great. Thank you!

mikibonacci:
Hi @danielhollas ! Thank you for this PR. 
In trying the new image, I have memory issues: when I try to open the App Store, the kernel dies. 
Maybe my laptop is not powerful enough (8 GB RAM), but this does not happen with the old full-stack image, so I it seems there is some additional memory consumption.

Not sure why, but I also tried rabbitmq-server==3.13.4 and the issue is still there. I will check more deeply, but it seems that when I use the notebook the memory usage increases from ~500 MB to ~700 MB (instead, with the old image, the usage remains constant at ~500 MB). It is not much, but maybe is sufficient to let the kernel die.

This is a screenshot of the stats of the container with the new image:

Screenshot 2024-07-20 at 13 29 52

Instead, the container with the old image show this stats:

Screenshot 2024-07-20 at 13 30 01

danielhollas:
@mikibonacci thank you very much for testing and debugging! 

Very strange, rabbitmq version should not be influencing the notebook in any way. So to understand, the AiiDAlab home page loads, but when you open a second page (app store) the kernel crashes?

I'll test locally as well. 

8Gb should be plenty. Can I ask how do you launch the container? It's possible that Docker ion Mac by default reduces available memory in the container?

mikibonacci:
I launched the container from the docker desktop interface, but also manually (not with aiidalab-launch), and for both the cases (new and old image). After running the command `rabbitmq-diagnostics status` (in the aiida-core-services env), I see that the total memory usage for the new-image container is 0.1 GB versus 0.0919 GB for the old. 

So maybe is not even a rabbitmq problem

danielhollas:
I can reproduce this on my computer, so this is not ARM specific, and probably comes from the new version of RabbitMQ. When I run `verdi status` it says it cannot connect to rabbitmq.

danielhollas:
Here's the error I see when running `rabbitmq-server start`

```
=INFO REPORT==== 20-Jul-2024::15:17:09.629468 ===
    alarm_handler: {set,{system_memory_high_watermark,[]}}
2024-07-20 15:17:09.785577+00:00 [warning] <0.156.0> Using RABBITMQ_ADVANCED_CONFIG_FILE: /opt/conda/envs/aiida-core-services/etc/rabbitmq/advanced.config
2024-07-20 15:17:10.657751+00:00 [error] <0.254.0> Feature flags: `classic_mirrored_queue_version`: required feature flag not enabled! It must be enabled before upgrading RabbitMQ.
2024-07-20 15:17:10.657800+00:00 [error] <0.254.0> Failed to initialize feature flags registry: {disabled_required_feature_flag,
2024-07-20 15:17:10.657800+00:00 [error] <0.254.0>                                               classic_mirrored_queue_version}
2024-07-20 15:17:10.662211+00:00 [error] <0.254.0> 
2024-07-20 15:17:10.662211+00:00 [error] <0.254.0> BOOT FAILED
2024-07-20 15:17:10.662211+00:00 [error] <0.254.0> ===========
2024-07-20 15:17:10.662211+00:00 [error] <0.254.0> Error during startup: {error,failed_to_initialize_feature_flags_registry}
2024-07-20 15:17:10.662211+00:00 [error] <0.254.0> 

```

danielhollas:
Looks like there might be multiple issues:

- If upgrading from a previously run container, rabbitmq may fail to start .Looks like upgrading RabbitMQ to from 3.9 directly to 3.13 might not be possible. 
https://www.rabbitmq.com/docs/upgrade#rolling-upgrades
- Appstore page crashing seems to be independent from whether rabbitmq is running or not (as it should). I have no idea why it is crushing on this PR though, will need to investigate more.

danielhollas:
I have filled a bug report about the appstore crashing: https://github.com/aiidalab/aiidalab/issues/442 This issue is also present on `aiidalab/full-stack:edge` and is unrelated to this PR.

@mikibonacci  I've rebased this PR and now the AppStore should not be crashing. But in any case, you can also install QEApp via command line with `aiidalab install quantum-espresso`, and that should allow you to test it. Thank you!

mikibonacci (approved):
Hi @danielhollas , I tested for relax + band structure calculation and everything was fine. For me it can be merged. Thanks!

build.json
```diff
@@ -7,13 +7,13 @@
       "default": "15"
     },
     "RMQ_VERSION": {
-      "default": "3.9.13"
+      "default": "3.13.5"
```
unkcpz, line 10: Hi @danielhollas, you mentioned once you want to bring this up to aiida team, I was on holiday that week. @superstar54 told me you were there at the meeting and mentioned this. Is there going to be some tests add for the different version of RMQ?

stack/base-with-services/Dockerfile
```diff
@@ -6,44 +6,20 @@ LABEL maintainer="AiiDAlab Team <aiidalab@materialscloud.org>"
...
-     rm rabbitmq-server-generic-unix-${RMQ_VERSION}.tar.xz && \
-     mv rabbitmq_server-${RMQ_VERSION} /opt/conda/envs/aiida-core-services/rabbitmq_server && \
-     ln -sf /opt/conda/envs/aiida-core-services/rabbitmq_server/sbin/* /opt/conda/envs/aiida-core-services/bin/ && \
-     fix-permissions "${CONDA_DIR}"; \
-  fi
+RUN mamba create -p /opt/conda/envs/aiida-core-services --yes \
+        postgresql=${PGSQL_VERSION} \
+        rabbitmq-server=${RMQ_VERSION} && \
+    mamba clean --all -f -y && \
+    fix-permissions "${CONDA_DIR}"
```
unkcpz, line 22: Cool!

unkcpz (approved):
Looks all good, I don't have too much to test. Since it only affect arm64, I'll trust @mikibonacci's tests.

danielhollas:
Thank you for testing @mikibonacci. 

Unfortunately, this cannot be merged yet, since it seems that when updating the image in an existing deployment that used the older RMQ version, the RMQ server fails to start, see my previous message with a link about rmq upgrades. We need to figure out how to upgrade safely first.

unkcpz:
For the update, if we can assure daemon is off is that safe to delete old RMQ configs in home entirely and start from newly installed one?

danielhollas:
> For the update, if we can assure daemon is off is that safe to delete old RMQ configs in home entirely and start from newly installed one?

Yeah, that was my thought as well, but I am not sure what will happen if for example user had some calculations running, and then turned off the container. If we then delete RMQ config, can this loose some messages? What would be the effect of that? I don't understand the details well enough, maybe something to be brought up at aiida-core meeting, maybe Seb would have a better idea.

danielhollas:
@yakutovicha I looked briefly into the `~/.rabbitmq` but don't see a way to determine RMQ version from the files present. If you know of a method please let me know.

That being said, I am hopefully (:crossed_fingers:) going to be able to publish RMQ 3.10.x version with arm64 build  on conda-forge sometimes next week, which would allow us to bypass this upgrade problem for now. Relevant PRs:

https://github.com/conda-forge/erlang-feedstock/pull/177
https://github.com/conda-forge/rabbitmq-server-feedstock/pull/127

danielhollas:
> That being said, I am hopefully (:crossed_fingers:) going to be able to publish RMQ 3.10.x version with arm64 build on conda-forge sometimes next week, which would allow us to bypass this upgrade problem for now. Relevant PRs:
https://github.com/conda-forge/erlang-feedstock/pull/177
https://github.com/conda-forge/rabbitmq-server-feedstock/pull/127

I made a lot of progress here over the weekend and merged the erlang PR (which was the major blocker). Unfortunately, due to a conda-forge outage, the new erlang package is not yet available for installation, so I cannot test and merge the corresponding RabbitMQ PR. Hopefully will be able to do this by the end of this week.

https://conda-forge.org/status/
image

danielhollas:
I am going to do a bit of testing, but otherwise this is ready! :tada: 

Ideally somebody with a Mac machine would test this as well.

edan-bainglass:
> I am going to do a bit of testing, but otherwise this is ready! 🎉
> 
> Ideally somebody with a Mac machine would test this as well.

Pinging @mbercx @giovannipizzi 🙏

mbercx:
Sure, happy to test. So I don't have to read the entire context: what should I test, specifically?

danielhollas:
Thanks @mbercx! You can pull the image from this PR with 

```
docker pull ghcr.io/aiidalab/full-stack:pr-482
```

First thing to test is whether the RMQ messaging works at all, e.g. just running some simple calculation in QeApp should do the trick.

Secondly, we want to ensure that the users can migrate from the previous image with RMQ 3.9 to this one with RMQ 3.10 (since we saw that e.g. upgrading to 3.11 causes problems). So for that you want to start from fresh container from a previous image, e.g. `ghcr.io/aiidalab/full-stack:edge`, run some calculations (to make sure that RMQ has been initialized with the old version), stop the container, switch the image to pr-482, and relaunch (using the same home volume).

LMK in case the above doesn't make sense. I did this test on my linux x86 machine and all looks good.

mbercx:
Thanks for the explanation @danielhollas! I'm currently on a train with poor connection, will give it a try tomorrow. :)

danielhollas:
Pitty, I really wanted you to work on this on Sunday! 😛

mbercx:
So, the calculation seems to run fine, but the `aiida.out` preview just shows the following:

Screenshot 2026-04-27 at 13 03 48

This is just using the image here, no update from `edge`.

I'm afraid I do not have time to do a lot of investigating / debugging here... Maybe I can give @edan-bainglass access to my mac Studio at the office, so he has a testing environment?

danielhollas:
Thank you! I saw this as well in my app which uses the same widget to
display the output from the QM calculation. Would you mind please
confirming my hypothesis and try to reproduce the problem with the edge
image? Then you'll be off the hook here :-)

I think opening an issue on QeApp is likely correct. I suspect this might
be caused by aiida-core 2.8. but I also might be totally wrong here

mbercx:
> Would you mind please confirming my hypothesis and try to reproduce the problem with the edge image? Then you'll be off the hook here :-)

Can confirm! Have the same issue on `edge`. 

Can I go out and play with my friends now? 🥹

danielhollas:
> Can I go out and play with my friends now? 🥹

Yes, but be back home by 10 o'clock!

danielhollas:
Okay, I opened https://github.com/aiidalab/aiidalab-qe/issues/1491 with the issue Marnik described above. Since it is unrelated to RMQ I am going to go ahead and merge this. Thank you all for testing!

PR #483, opened 2024-07-20
Preinstall AWB in lab image

### LINKED ISSUE #526: scipy from conda needed for correct BLAS/LAPACK operations in arm64
This is a problem related to the QE app in `arm64`, in particular with the `aiidalab-qe-vibroscopy` plugin. The problem is in the calculation of inelastic neutron scattering spectra, which are computed in euphonic using `scipy` modules, which are using BLAS/LAPACK. However, in `arm64` there is some broken libraries/linking if `scipy` is installed from `pip`, and we need it from `conda`. Otherwise, when INS data are processed, the app gets stuck and cannot compute spectra. 

so, we should add `conda install scipy==1.13.1 --y` somewhere (or the `mamba` analogous). 

I already added this in the `aiidalab/qe` image, but we should do it here as the user can install the app also just using the `aiidalab/full-stack` image. 

As final comment, to be safe and cover also corner cases in non `arm64` machines, I suggest we always install `scipy` via `conda`.

danielhollas:
The goal here is to pre-install AWB into the image from a conda package to:

- speed up the installation of apps that depend on AWB
- make the image more robust by pre-installing the often-used  dependencies into the conda environment so that they don't need to be installed from PyPI by pip into `~/.local`
- relatedly, this PR also proposes to pre-install the `atomic_tools` extras from aiida-core which share a lot of the packages with AWB (e.g. ase) and so it is generally good to ensure that the supported versions are compatible with each other. Also, `atomic_tools` includes pymatgen and scipy, which we had issues with installing from PyPI before (see #526)

Obviously, the big disadvantage is the size of the image, but to me the increase stability seems worth it. Also, applications dependending on AWB (like QeApp) would end up installing the packages anyways, and installing them from conda is actually generally more efficient since conda packages do not vendor non-python code).

NOTE: We cannot **pin** the exact AWB version, because apps need to be able to upgrade or even downgrade the pre-installed version according to their needs. But the thesis is that most of the AWB dependencies are not changed in between AWB versions and so it still makes sense to have them pre-installed. We probably can specify the lowest supported AWB version (which since this is build on the python 3.12 image will be AWB 3.0)

Staged on top of #455

Closes #526

### DIFF
--- a/stack/lab/Dockerfile
+++ b/stack/lab/Dockerfile
@@ -11,11 +11,12 @@ ENV DOCKER_STACKS_JUPYTER_CMD=nbclassic
 USER root
 WORKDIR /opt/
 
-# Install aiidalab package
 ARG AIIDALAB_VERSION
+ARG AWB_VERSION
 RUN mamba install --yes \
      aiidalab=${AIIDALAB_VERSION} \
      appmode=1.1.0 \
+     aiidalab-widgets-base=${AWB_VERSION} \
      && mamba clean --all -f -y && \
      fix-permissions "${CONDA_DIR}" && \
      fix-permissions "/home/${NB_USER}"
--- a/build.json
+++ b/build.json
@@ -15,6 +15,9 @@
     "AIIDALAB_VERSION": {
       "default": "26.06.0"
     },
+    "AWB_VERSION": {
+      "default": "3.0.0a2"
+    },
     "AIIDALAB_HOME_TAG": {
       "default": "v26.06.0"
     }
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -17,6 +17,9 @@ variable "AIIDA_VERSION" {
 variable "AIIDALAB_VERSION" {
 }
 
+variable "AWB_VERSION" {
+}
+
 variable "AIIDALAB_HOME_TAG" {
 }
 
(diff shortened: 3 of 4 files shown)


danielhollas:
For some reason ASE on conda-forge depends on flask, which bring in a ton of other dependencies. I opened a PR on the ASE feedstock to remove it
https://github.com/conda-forge/ase-feedstock/pull/59

stack/lab/Dockerfile
```diff
@@ -11,11 +11,11 @@ ENV DOCKER_STACKS_JUPYTER_CMD=nbclassic
 USER root
 WORKDIR /opt/
 
-# Install aiidalab package
 ARG AIIDALAB_VERSION
+ARG AWB_VERSION
 RUN mamba install --yes \
      aiidalab=${AIIDALAB_VERSION} \
-     appmode=1.1.0 \
+     aiidalab-widgets-base=${AWB_VERSION} \
```
danielhollas, line 19:
```suggestion
     appmode=1.1.0 \
     aiidalab-widgets-base=${AWB_VERSION} \
```

danielhollas:
Faling tests are because of `pip check`

> E           Exception: Command docker compose -f "stack/docker-compose.lab.yml" -p "pytest2074" exec -T aiidalab pip check returned 1: """aiidalab-widgets-base 3.0.0a0 has requirement nglview>=3.0.8,~=3.0, but you have nglview 0.0.0.
>           """.
> /opt/hostedtoolcache/Python/3.12.13/arm64/lib/python3.12/site-packages/pytest_docker/plugin.py:37: Exception


Looks like bad metadata of `nglview` package on conda-forge. This should be fixed if we can update to nglview 4.0 (https://github.com/aiidalab/aiidalab-widgets-base/pull/742)

edan-bainglass (approved):
LGTM! Thanks @danielhollas. When you say "the big disadvantage is the size of the image" - how big?

danielhollas:
Thanks @edan-bainglass 

(this is staged on top of #479 so not merging yet). Also would like to get green light from @yakutovicha

yakutovicha (approved):
I like it!

PR #485, opened 2024-07-22
Add pip.conf

### LINKED ISSUE #484: Fix `pip list` behaviour
In #437, we've set the global environment variable `PIP_USER=1`, which means that `pip install` would by default install to user site in `~/.local` (i.e. `pip install` would become `pip install --user`. 

What I did not realize is that this would affect other pip commands that have `--user` option. One such command is `pip list`, which we use in `aiidalab` to list installed packages. Unfortunately, `pip list --user` will only list packages in `~/.local`, and will thus miss all the packages installed in the conda environment.

To fix this, instead of using `PIP_USER` envvar, we can create a pip configuration file, and set a per-command section so that `--user` is set only for `pip install`.  https://pip.pypa.io/en/stable/topics/configuration/#per-command-section

I've noticed this when playing with the QeApp image, where I would see this on the app page:
image

danielhollas:
Follow up to #437. Fixes #484 (see this issue for more details)

PIP_USER envvar affects all pip subcommands, including `pip list`  which had unintended side effects. Using the pip.conf file allows to set `--user` specifically for the `install` subcommand.

### DIFF
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -42,7 +42,7 @@ RUN cat /opt/requirements.txt | xargs -I{} conda config --system --add pinned_pa
 # Configure pip to use the same requirements file as constraints file.
 ENV PIP_CONSTRAINT /opt/requirements.txt
 # Ensure that pip installs packages to ~/.local by default
-ENV PIP_USER 1
+COPY pip.conf /etc/pip.conf
 
 # Upgrade pip and mamba to latest
 # Install aiida-core and other shared requirements.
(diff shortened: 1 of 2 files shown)


unkcpz (approved):
It's been months I didn't play with QEApp. Thanks for the fix! @danielhollas 

@superstar54 I'd recommend to have a read on https://github.com/aiidalab/aiidalab-docker-stack/issues/484. It may nicely answer the question you had during our meeting with Daniel about two places for package installation.

PR #489, opened 2024-08-02
Fix `pip check` test

### LINKED ISSUE #490: Latest pip `pip check` fails for several packages
This is what I see running `pip check` on new builds:

```
async-generator 1.10 is not supported on this platform
certipy 0.1.3 is not supported on this platform
upf-to-json 0.9.2 is not supported on this platform
```

This is a consequence of https://github.com/pypa/pip/issues/12884

~~Short term solution is probably to pin pip to previous version.~~ EDIT: I've fixed the packages on conda-forge instead.

danielhollas:
Fixes #490. The offending packages have been fixed upstream, but we need to forcibly update `async_generator` and `certipy` to pick up those changes since these are already part of the jupyter base image.

Additionally, I've unpinned the `softprops/actions-gh-release` action.
Pinning this action to specific patch version just creates noise from dependabot. Given that we currently don't pin any other actions to the exact versions, let's unpin this one as well.

Closes #487

### DIFF
--- a/stack/base/Dockerfile
+++ b/stack/base/Dockerfile
@@ -45,14 +45,15 @@ ENV PIP_CONSTRAINT /opt/requirements.txt
 COPY pip.conf /etc/pip.conf
 
 # Upgrade pip and mamba to latest
+# Update async_generator, certipy to satisfy `pip check`
+# https://github.com/aiidalab/aiidalab-docker-stack/issues/490
 # Install aiida-core and other shared requirements.
-RUN mamba update -y pip && \
+RUN mamba update -y pip async_generator certipy && \
      mamba install --yes \
      aiida-core==${AIIDA_VERSION} \
      mamba-bash-completion \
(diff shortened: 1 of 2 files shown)


stack/base/Dockerfile
```diff
@@ -45,14 +45,15 @@ ENV PIP_CONSTRAINT /opt/requirements.txt
...
+# https://github.com/aiidalab/aiidalab-docker-stack/issues/490
 # Install aiida-core and other shared requirements.
-RUN mamba update -y pip && \
+RUN mamba update -y pip async_generator certipy && \
      mamba install --yes \
      aiida-core==${AIIDA_VERSION} \
      mamba-bash-completion \
      && mamba clean --all -f -y && \
-     fix-permissions "${CONDA_DIR}" && \
-     fix-permissions "/home/${NB_USER}"
```
danielhollas, line 55: This command should not be touching home folder at all so we can simply remove this.

PR #492, opened 2024-08-07
Pin docker/bake-action to fix build

### LINKED ISSUE #491: Failing build
The latest build on main failed with a rather cryptic error in `Set output variables` step which runs `extract-image-names.sh`.

```
Error: An error occurred trying to start process '/usr/bin/bash' with working directory '/home/runner/work/aiidalab-docker-stack/aiidalab-docker-stack'. Argument list too long
```
https://github.com/aiidalab/aiidalab-docker-stack/actions/runs/10283635135/job/28465139810

It looks like the metadata output from docker/bake-action has gotten too big.

danielhollas:
Fixes #491

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,7 +55,9 @@ jobs:
 
             - name: Build and upload to ghcr.io 📤
               id: build-upload
-              uses: docker/bake-action@v5
+              # We need to pin exact version here, since updates can break
+              # the extract-image-names.sh script
+              uses: docker/bake-action@v5.5.0
               with:
                   push: true
                   # Using provenance to disable default attestation so it will build only desired images:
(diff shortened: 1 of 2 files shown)


danielhollas:
I've been able to workaround the original issue detailed in #491, but than ran into another one since the specification of the metadata json changes and the `jq` script in `extract-image-names.sh` no longer works. So for now I've pinned the `bake-action` to the previous version. I have nevertheless kept the workaround/fix since I think it is better then the original approach.

unkcpz:
Parsing things with `jq` is already quite complex to me. I am interested to know how you debug it?

danielhollas:
I am also unfamiliar with jq. The original error was caused by bash. Once I fixed it I ran into a jq error https://github.com/aiidalab/aiidalab-docker-stack/actions/runs/10287638235/job/28471150939

```
jq: error (at bake_metadata.json:691): Cannot index array with string "image.name"
```

I have verified that the json schema indeed changed, but I did not debug further and opened #493. I think the fix will be simple for somebody who knows `jq`. :sweat_smile:

PR #498, opened 2024-09-16
Move daemon start after so more computer setup can be added by children images

unkcpz:
In QeApp image, I need to stop daemon, setup computer and restart daemon. I seems it can be optimized by postpone the starting daemon here.


unkcpz:
replaced by #500

PR #500, opened 2024-09-16
Move daemon start after so more computer setup can be added by children images

unkcpz:


stack/base/before-notebook.d/99_start_aiida_daemon.sh
danielhollas, line 1: Wonder if this should be renamed to `90_start_aiida_daemon.sh` in case we have some scripts that rely on the running deamon in the future?

danielhollas (approved):
LGMT!

stack/base/before-notebook.d/99_start_aiida_daemon.sh
unkcpz: Good idea! I changed.

PR #510, opened 2025-01-06
Update RMQ to 3.10.25

danielhollas:
For now I'm just testing if we can straightforwardly update to 3.10 from 3.9

See:
https://www.rabbitmq.com/docs/which-erlang#old-timers
https://www.rabbitmq.com/docs/upgrade#rabbitmq-version-upgradability

TODO:
 - The arm64 build will be broken here since the system erlang will be incompatible

### DIFF
--- a/build.json
+++ b/build.json
@@ -7,7 +7,7 @@
       "default": "15"
     },
     "RMQ_VERSION": {
-      "default": "3.9.13"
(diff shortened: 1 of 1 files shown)


danielhollas:
Closing in favour of #482

PR #514, opened 2025-01-17
Switch to GitHub Arm runners

danielhollas:
Let's see how this works!

https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/


danielhollas:
It looks like this works! We could now easily run arm tests on all PRs and not worry about buildjet costs. The timings look comparable, perhaps even slightly better on GitHub!

unkcpz:
wow, this is super nice, we wait this for more than a year.

PR #516, opened 2025-01-22
Pre-compile requirements.txt with uv pip compile

danielhollas:
To make our CI more robust, we pin all the Python dependencies to a specific version, using `uv pip compile requirements.in > requirements.txt`.

We take advantage of uv's universal resolver so the generated `requirements.txt` file should work for all python versions >3.9 and all platforms (linux, mac, windows).


edan-bainglass (approved):
LGTM!

PR #522, opened 2025-02-13
Fix arm build

danielhollas:
Downgrading the QEMU version seems to fix the ARM build so let's go with that. Fix suggested in https://github.com/tonistiigi/binfmt/issues/215#issuecomment-2654328833


edan-bainglass (approved):
Let's do it! Thanks @danielhollas

PR #524, opened 2025-04-01
Implement opt-in countdown feature and add to image recipe

edan-bainglass:
This PR reimplemented [aiidalab-timer](https://github.com/edan-bainglass/aiidalab-timer) as a core feature of AiiDAlab instead of an app.

---

Jupyter provides entry points to customize notebook JS and CSS (see [here](https://jupyter-notebook.readthedocs.io/en/4.x/examples/Notebook/rstversions/JavaScript%20Notebook%20Extensions.html#custom-js) for old, but as far as I can tell, still valid docs). Following this approach, the PR introduces under `lab/countdown/` the following:
- `custom.js` - a script implementing the countdown timer, applying it consistently throughout AiiDAlab
- `custom.css` - styling the countdown-related DOM as a sticky banner above the native navbar
- `config.json.template` - a template with an opt-in toggle and a lifetime variable, to be read by `custom.js`

The PR also introduces a new script under `before_notebook.d` that runs on container creation and replaces `config.json.template` with `config.json`, applying the `EPHEMERAL` and `EXPIRY` environment variables, the latter computed from the opt-in `LIFETIME` environment variable (used by each deployment), and the former set to `true` if `LIFETIME` is present, `false` otherwise. This allows each deployment to opt-in and choose its own expiration.

Lastly, the PR updates the `lab/Dockerfile` to copy the necessary files to `site-packages/notebook/static/custom/` and add the `PYTHON_MINOR_VERSION` environment variable required by the container-runtime script.

---
image
---
image
---
image

### DIFF
--- a/stack/lab/Dockerfile
+++ b/stack/lab/Dockerfile
@@ -47,6 +47,11 @@ RUN jupyter nbextension enable --py --sys-prefix appmode && \
 # Swap appmode icon for AiiDAlab gears icon, shown during app load
 COPY --chown=${NB_UID}:${NB_GID} gears.svg ${CONDA_DIR}/share/jupyter/nbextensions/appmode/gears.svg
 
+# Set up opt-in countdown feature
+ARG PYTHON_MINOR_VERSION
+ENV PYTHON_MINOR_VERSION=${PYTHON_MINOR_VERSION}
+COPY --chown=${NB_UID}:${NB_GID} countdown/ ${CONDA_DIR}/lib/python${PYTHON_MINOR_VERSION}/site-packages/notebook/static/custom/
+
 # Copy start-up scripts for AiiDAlab.
 COPY before-notebook.d/* /usr/local/bin/before-notebook.d/
 
@@ -99,5 +104,4 @@ ENV NOTEBOOK_ARGS=\
 "--TerminalManager.cull_interval=300"
 
 # Set up the logo of notebook interface
-ARG PYTHON_MINOR_VERSION
 COPY --chown=${NB_UID}:${NB_GID} aiidalab-wide-logo.png ${CONDA_DIR}/lib/python${PYTHON_MINOR_VERSION}/site-packages/notebook/static/base/images/logo.png
--- a/stack/lab/before-notebook.d/70_prepare_countdown_config.sh
+++ b/stack/lab/before-notebook.d/70_prepare_countdown_config.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+set -e
+
+CUSTOM_DIR="${CONDA_DIR}/lib/python${PYTHON_MINOR_VERSION}/site-packages/notebook/static/custom"
+
+if [ "$LIFETIME" ]; then
+    EPHEMERAL=true
+
+    # Convert LIFETIME from HH:MM:SS to seconds
+    IFS=: read -r H M S <<<"$LIFETIME"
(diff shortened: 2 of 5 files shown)


stack/lab/Dockerfile
```diff
@@ -47,6 +47,11 @@ RUN jupyter nbextension enable --py --sys-prefix appmode && \
 # Swap appmode icon for AiiDAlab gears icon, shown during app load
 COPY --chown=${NB_UID}:${NB_GID} gears.svg ${CONDA_DIR}/share/jupyter/nbextensions/appmode/gears.svg
 
+# Set up opt-in countdown feature
+ARG PYTHON_MINOR_VERSION
+ENV PYTHON_MINOR_VERSION=${PYTHON_MINOR_VERSION}
+COPY --chown=${NB_UID}:${NB_GID} countdown/ ${CONDA_DIR}/lib/python${PYTHON_MINOR_VERSION}/site-packages/notebook/static/custom/
+
```
unkcpz, line 54: These lines were moved to here when we debug it together. Can you move it to where the logo.png is copied, and see if it still work?

edan-bainglass:
Unclear as to why the test is failing. Seems unrelated, but surely I'm missing something.

danielhollas:
> Unclear as to why the test is failing. Seems unrelated, but surely I'm missing something.

Please reopen the PR from origin, otherwise you won't have permissions to push to ghcr.io

unkcpz (changes requested):
let's wait a bit to see if there are standard way to get the remaining cull time, I asked in https://discourse.jupyter.org/t/how-to-get-the-remaining-cull-time-for-the-running-container/34244

I also not so sure this timeout should put in the aiidalab-docker-stack. Is the timeout only for qeapp? If so, probably the better solution is to set a k8s timeout there instead of using the cull time from jupyter server. 
To make the container closed at exactly after it start 12h during the deployment to k8s I believe we can use `activeDeadlineSeconds` https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup

With that approach there is not much need to change with this PR, the JS/css still fit and can be used for deployment when the timeout is set. 

I didn't review the js part (I can read but not qualified to review, and I refuse to learn js :sunglasses: ), css part looks good.

stack/lab/Dockerfile
edan-bainglass: It doesn't work due to the order of events I believe

edan-bainglass:
> let's wait a bit to see if there are standard way to get the remaining cull time, I asked in https://discourse.jupyter.org/t/how-to-get-the-remaining-cull-time-for-the-running-container/34244
> 
> I also not so sure this timeout should put in the aiidalab-docker-stack. Is the timeout only for qeapp? If so, probably the better solution is to set a k8s timeout there instead of using the cull time from jupyter server. To make the container closed at exactly after it start 12h during the deployment to k8s I believe we can use `activeDeadlineSeconds` https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup
> 
> With that approach there is not much need to change with this PR, the JS/css still fit and can be used for deployment when the timeout is set.
> 
> I didn't review the js part (I can read but not qualified to review, and I refuse to learn js 😎 ), css part looks good.

I'm not sure I understand your concerns. Let's discuss in person.

edan-bainglass:
Relocating due to permissions

PR #525, opened 2025-04-01
Implement opt-in countdown feature and add to image recipe

edan-bainglass:
This PR reimplemented [aiidalab-timer](https://github.com/edan-bainglass/aiidalab-timer) as a core feature of AiiDAlab instead of an app.

---

Jupyter provides entry points to customize notebook JS and CSS (see [here](https://jupyter-notebook.readthedocs.io/en/4.x/examples/Notebook/rstversions/JavaScript%20Notebook%20Extensions.html#custom-js) for old, but as far as I can tell, still valid docs). Following this approach, the PR introduces under `lab/countdown/` the following:
- `custom.js` - a script implementing the countdown timer, applying it consistently throughout AiiDAlab
- `custom.css` - styling the countdown-related DOM as a sticky banner above the native navbar
- `config.json.template` - a template with an opt-in toggle and a lifetime variable, to be read by `custom.js`

The PR also introduces a new script under `before_notebook.d` that runs on container creation and replaces `config.json.template` with `config.json`, applying the `EPHEMERAL` and `EXPIRY` environment variables, the latter computed from the opt-in `LIFETIME` environment variable (used by each deployment), and the former set to `true` if `LIFETIME` is present, `false` otherwise. This allows each deployment to opt-in and choose its own expiration.

Lastly, the PR updates the `lab/Dockerfile` to copy the necessary files to `site-packages/notebook/static/custom/` and add the `PYTHON_MINOR_VERSION` environment variable required by the container-runtime script.

---
image
---
image
---
image

### DIFF
--- a/stack/lab/Dockerfile
+++ b/stack/lab/Dockerfile
@@ -47,6 +47,11 @@ RUN jupyter nbextension enable --py --sys-prefix appmode && \
 # Swap appmode icon for AiiDAlab gears icon, shown during app load
 COPY --chown=${NB_UID}:${NB_GID} gears.svg ${CONDA_DIR}/share/jupyter/nbextensions/appmode/gears.svg
 
+# Set up opt-in countdown feature
+ARG PYTHON_MINOR_VERSION
+ENV PYTHON_MINOR_VERSION=${PYTHON_MINOR_VERSION}
+COPY --chown=${NB_UID}:${NB_GID} countdown/ ${CONDA_DIR}/lib/python${PYTHON_MINOR_VERSION}/site-packages/notebook/static/custom/
+
 # Copy start-up scripts for AiiDAlab.
 COPY before-notebook.d/* /usr/local/bin/before-notebook.d/
 
@@ -99,5 +104,4 @@ ENV NOTEBOOK_ARGS=\
 "--TerminalManager.cull_interval=300"
 
 # Set up the logo of notebook interface
-ARG PYTHON_MINOR_VERSION
 COPY --chown=${NB_UID}:${NB_GID} aiidalab-wide-logo.png ${CONDA_DIR}/lib/python${PYTHON_MINOR_VERSION}/site-packages/notebook/static/base/images/logo.png
--- a/stack/lab/before-notebook.d/70_prepare_countdown_config.sh
+++ b/stack/lab/before-notebook.d/70_prepare_countdown_config.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+set -e
+
+CUSTOM_DIR="${CONDA_DIR}/lib/python${PYTHON_MINOR_VERSION}/site-packages/notebook/static/custom"
+
+if [ "$LIFETIME" ]; then
+    EPHEMERAL=true
+
+    # Convert LIFETIME from HH:MM:SS to seconds
+    IFS=: read -r H M S <<<"$LIFETIME"
+    LIFETIME_SEC=$((10#$H * 3600 + 10#$M * 60 + 10#$S))
+
+    # Calculate expiry timestamp in UTC
+    EXPIRY=$(date -u -d "+${LIFETIME_SEC} seconds" +"%Y-%m-%dT%H:%M:%SZ")
+    export EXPIRY
+else
+    EPHEMERAL=false
+fi
+
+export EPHEMERAL
+envsubst <"${CUSTOM_DIR}/config.json.template" >"$CUSTOM_DIR/config.json"
+rm "${CUSTOM_DIR}/config.json.template"
--- a/stack/lab/countdown/custom.js
+++ b/stack/lab/countdown/custom.js
@@ -1,0 +1,103 @@
+require(["base/js/namespace", "base/js/events"], (Jupyter, events) => {
+  const parseLifetimeToMs = (str) => {
+    const parts = str.split(":").map(Number);
+    if (parts.length !== 3 || parts.some(isNaN)) {
+      return null;
+    }
+    const [h, m, s] = parts;
+    return ((h * 60 + m) * 60 + s) * 1000;
+  };
+
+  const insertCountdown = (remainingMs) => {
+    if (document.getElementById("culling-countdown")) {
+      return;
+    }
+
+    const banner = document.createElement("div");
+    banner.id = "culling-countdown";
+
+    const shutdownWarning = document.createElement("div");
+    shutdownWarning.id = "shutdown-warning";
+    shutdownWarning.innerHTML = "⚠️ Shutdown imminent! ⚠️";
+    banner.appendChild(shutdownWarning);
+
+    const countdown = document.createElement("div");
+    countdown.id = "countdown";
+    countdown.innerHTML = `Session time remaining: `;
+    const timer = document.createElement("span");
+    timer.id = "culling-timer";
+    timer.innerHTML = "Calculating...";
+    countdown.appendChild(timer);
+    banner.appendChild(countdown);
+
+    const saveInfo = document.createElement("div");
+    saveInfo.id = "save-info";
+    saveInfo.innerHTML = `
+      Consider saving your work using the <b>File Manager</b> or the <b>Terminal</b>
+    `;
+    banner.appendChild(saveInfo);
+
+    const endTime = new Date(Date.now() + remainingMs);
+
+    const formatTime = (seconds) => {
+      const hrs = `${Math.floor(seconds / 3600)}`.padStart(2, "0");
+      const mins = `${Math.floor((seconds % 3600) / 60)}`.padStart(2, "0");
+      const secs = `${Math.floor(seconds % 60)}`.padStart(2, "0");
+      return `${hrs}:${mins}:${secs}`;
+    };
+
+    const updateTimer = () => {
+      const now = new Date();
+      const timeLeft = (endTime - now) / 1000;
+      if (timeLeft < 0) {
+        clearInterval(interval);
+        return;
+      }
+      if (timeLeft < 1800) {
+        banner.style.backgroundColor = "#DAA801";
+        saveInfo.style.display = "block";
+      }
+      if (timeLeft < 300) {
+        banner.style.backgroundColor = "red";
+        shutdownWarning.style.display = "block";
+        shutdownWarning.innerHTML = "⚠️ Shutdown imminent ⚠️";
+      }
+      timer.innerHTML = formatTime(timeLeft);
+    };
+
+    updateTimer();
+    const interval = setInterval(updateTimer, 1000);
+
+    const container = document.getElementById("header");
+    if (container) {
+      container.parentNode.insertBefore(banner, container);
+    }
+  };
+
+  loadCountdown = async () => {
+    try {
+      const response = await fetch("/static/custom/config.json");
+      const config = await response.json();
+
+      // Opt-in point for deployments
+      if (!config.ephemeral) {
+        return;
+      }
+
+      if (!config.expiry) {
+        console.warn("Missing `expiry` in config file");
+        return;
+      }
+
+      const expiry = new Date(config.expiry).getTime();
+      const remaining = expiry - Date.now();
+
+      insertCountdown(Math.max(0, remaining));
+    } catch (err) {
+      console.error("Countdown init failed:", err);
+    }
+  };
+
+  events.on("app_initialized.NotebookApp", loadCountdown);
+  loadCountdown();
+});
--- a/stack/lab/countdown/config.json.template
+++ b/stack/lab/countdown/config.json.template
@@ -1,0 +1,4 @@
+{
+  "ephemeral": ${EPHEMERAL},
+  "expiry": "${EXPIRY}"
+}
--- a/stack/lab/countdown/custom.css
+++ b/stack/lab/countdown/custom.css
@@ -1,0 +1,33 @@
+#culling-countdown {
+  position: sticky;
+  top: 0;
+  left: 0;
+  width: 100%;
+  background-color: #0078d4;
+  color: white;
+  text-align: center;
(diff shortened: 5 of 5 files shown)


edan-bainglass:
@unkcpz the feature is now timezone-independent. However, it is a bit hacky. Let's discuss in the office.

edan-bainglass:
> @unkcpz the feature is now timezone-independent. However, it is a bit hacky. Maybe we can clean it up. Let's discuss in the office.

unkcpz:
About the culling time setup I mentioned in the #524, I talk to @edan-bainglass off-line and we agree if it is the rigid 12h no matter user interact with notebook or not, which means the timer shows the left time exactly how long the pod or kernel will be shutdown. Then I think it is a correct implementation. 

From the demo server deployment config in https://github.com/aiidalab/aiidalab-demo-server/blob/23f91a638588de9241dba8de6f905886efb5ddf4/basehub/values.yaml.j2#L100-L106. This apparently is the config for the `jupyterhub` helm chart, so I worried this set for the culling idle time? @yakutovicha can you double check? thanks.

The k8s when deploy can also set a rigid timeout for the pod which might be more reliable and you may want to use (https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup).

stack/lab/Dockerfile
```diff
@@ -47,6 +47,18 @@ RUN jupyter nbextension enable --py --sys-prefix appmode && \
...
+# We expose PYTHON_MINOR_VERSION in the container, as it is used by the
+# before-notebook.d/70_prepare_countdown_config.sh script to write the
+# environment-variable-dependent (opt-in) configuration file.
+ARG PYTHON_MINOR_VERSION
+ENV PYTHON_MINOR_VERSION=${PYTHON_MINOR_VERSION}
+COPY --chown=${NB_UID}:${NB_GID} countdown/ ${CONDA_DIR}/lib/python${PYTHON_MINOR_VERSION}/site-packages/notebook/static/custom/
+
+# Add endpoint to fetch container uptime
+COPY --chown=${NB_UID}:${NB_GID} aiidalab_uptime.py ${CONDA_DIR}/lib/python${PYTHON_MINOR_VERSION}/site-packages/
+RUN echo "c.NotebookApp.server_extensions.append('aiidalab_uptime')" >> /etc/jupyter/jupyter_notebook_config.py
```
unkcpz, line 60: As mentioned in the review of #524, can you move this part next to the logo.png copy? we move these lines here when we debug together and we found the problem was the env variable not properly set. 
Keep the PYTHON_MINOR_VERSION set to where it used would be easier for future maintenance.

unkcpz: The echo and pipeline to jupyter config looks very hacky but @edan-bainglass mentioned it is the only way it can work. 

If I understand correctly, you want to have an entry point for the https://<url>/uptime. I did this for template of the customize of jupyterhub index page in https://github.com/aiidalab/aiidalab-demo-server/blob/main/basehub/templates/hub-configmap.yaml
There is this config file you want to change https://github.com/aiidalab/aiidalab-demo-server/blob/main/basehub/files/etc/jupyterhub/jupyter_notebook_config.py.
Thus I think these changes need to moved there only for the demo server deployment.

unkcpz (changes requested):
I still think these changes can only stay in the demo-server deployment as I did for the customized jupyterhub index page through k8s ConfigMaps. Then the hacky solution could be avoid.

stack/lab/Dockerfile
edan-bainglass: A few things to clarify:
1. I mentioned it is the only way I could get it to work, as all other attempts failed. However, it is likely that I'm simply not following the correct approach, but the docs are lacking.
2. This feature is not only for the demo server deployment. It is a general feature we should offer to all "temporary" deployments
3. Indeed, one can adjust on the deployment side. One alternative to implementing this feature here would be to create a general deployment template and implement it there. This has been discussed, but not yet resolved.

That said, what is the issue with the feature living here? There is likely a cleaner solution to the `echo` solution.

unkcpz: > This feature is not only for the demo server deployment.

That's true, this count down thing can be generic for other deployment.

> That said, what is the issue with the feature living here?

I afraid even after this change when it deployment to the k8s, you still not see the timer there, since in our deployment, the `jupyter_conifg.py` is also override, I don't know which one will happened after and take effect. I think you did the correct change in this PR, but it requires more thoughts and test on the real k8s deployment and we decide where to put this changes. 

So my calls are:

- `stack/lab/before-notebook.d/70_prepare_countdown_config.sh` stay in this repo.
- `stack/lab/countdown/custom.css`stay in this repo.
- `stack/lab/countdown/custom.js` stay in this repo.
- `stack/lab/aiidalab_uptime.py` also stay here.
- only the changes in the docker file become the deployment specific thing and goes to demo-server repo and take effect there. Plus you need to add some documentation on how to use this count down thing if we want to make it generic for future customize deployment. thoughts?

edan-bainglass: Agreed. I'll dig some more into how we introduce the endpoint in a cleaner way. And docs for sure. Thanks @unkcpz 🙏

edan-bainglass: Solved. See description.

unkcpz:
@edan-bainglass my apologize that I am not able to give it a detail go in next two weeks. I am sick today and will start 2 weeks holiday next monday.

Since you notice the `jupyter_notebook_config.py` both in image and on the deployment, I think the changes should on the correct track. 
There are further doc that worth to read through (I believe you already did):
- https://jupyter-server.readthedocs.io/en/latest/users/configuration.html
- https://jupyter-notebook.readthedocs.io/en/v6.4.6/config.html

edan-bainglass:
> @edan-bainglass my apologize that I am not able to give it a detail go in next two weeks. I am sick today and will start 2 weeks holiday next monday.

Again, sorry to hear about your illness 😔

> Since you notice the `jupyter_notebook_config.py` both in image and on the deployment, I think the changes should on the correct track. There are further doc that worth to read through (I believe you already did):
> 
> * https://jupyter-server.readthedocs.io/en/latest/users/configuration.html
> * https://jupyter-notebook.readthedocs.io/en/v6.4.6/config.html

Indeed, I went through these (and others) in the process of deriving my solution.

@danielhollas I'll give you some time to review/comment, but great if we can get this out soon. The demo server is now public on the AiiDAlab website. Great if we can have this feature integrated 🙏

danielhollas:
I'll have a look today.

stack/lab/before-notebook.d/70_prepare_countdown_config.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+CUSTOM_DIR="${CONDA_DIR}/lib/python${PYTHON_MINOR_VERSION}/site-packages/notebook/static/custom"
+
+cat <<EOF >"${CUSTOM_DIR}/config.json"
+{
+  "ephemeral": $([ -n "$LIFETIME" ] && echo 1 || echo 0),
```
danielhollas, line 8: LIFETIME is quite generic. Perhaps `CONTAINER_LIFETIME` conveys its meaning better, if I understood the use case correctly?

danielhollas:
Hi @edan-bainglass,

this looks great, thanks for spending extra time on this. This is much closer what I imagined both in terms of UI and implementation. My feedback for the code is that I don't much like the custom lifetime parsing, but that can also be improved as a follow-up.

My main concern is similar to @unkcpz that I don't think this repo is the best place for this feature, at least for this first iteration. 

> That said, what is the issue with the feature living here? There is likely a cleaner solution to the echo solution.

The biggest downside is that this haven't been proven in production yet, and it's likely it will need tweaks or bugfixes. You'll be able to iterate much faster if all of this lives in the demo-stack repo or even the QeApp image.

While I agree that this can be a common feature, before we have an actual second use case, I'd prefer to not have it here. The docker stack here is already complex and quite painful to maintain, and this would add non-trivial complexity.

stack/lab/countdown/custom.js
```diff
@@ -0,0 +1,116 @@
...
+
+    const [_, days, hours, minutes, seconds] = match.map((v) => Number(v) || 0);
+
+    const totalSeconds = days * 86400 + hours * 3600 + minutes * 60 + seconds;
+    return totalSeconds * 1000;
+  };
+
+  const insertCountdown = (remainingMs) => {
+    if (document.getElementById("culling-countdown")) {
+      return;
```
danielhollas, line 20: Maybe we should log a warning here, since this should never happen if the CSS was loaded correctly?

stack/lab/countdown/custom.js
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    const [_, days, hours, minutes, seconds] = match.map((v) => Number(v) || 0);
+
+    const totalSeconds = days * 86400 + hours * 3600 + minutes * 60 + seconds;
+    return totalSeconds * 1000;
+  };
+
+  const insertCountdown = (remainingMs) => {
+    if (document.getElementById("culling-countdown")) {
```
danielhollas, line 19: Does `app_initialized.NotebookApp` event ensure that this CSS will be loaded by the time this function gets called?

stack/lab/countdown/custom.js
```diff
@@ -0,0 +1,116 @@
...
+    if (container) {
+      container.parentNode.insertBefore(banner, container);
+    }
+  };
+
+  const loadCountdown = async () => {
+    try {
+      const [configResponse, uptimeResponse] = await Promise.all([
+        fetch("/static/custom/config.json"),
+        fetch("/uptime"),
```
danielhollas, line 89: If I understand the logic correctly, this endpoint is called only once when the notebook is loaded, right?

edan-bainglass:
Fair. Will move to demo server repo until proven stable and/or required by other deployments. Thanks for the review @danielhollas and @unkcpz 🙏

danielhollas (changes requested):

stack/lab/Dockerfile
```diff
@@ -45,9 +45,25 @@ RUN jupyter nbextension enable --py --sys-prefix appmode && \
...
+ARG PYTHON_MINOR_VERSION
+
+# Set up the logo of notebook interface
+COPY aiidalab-wide-logo.png ${CONDA_DIR}/lib/python${PYTHON_MINOR_VERSION}/site-packages/notebook/static/base/images/logo.png
+
+# Copy custom CSS and JS files for AiiDAlab container countdown feature
+COPY countdown/ ${CONDA_DIR}/lib/python${PYTHON_MINOR_VERSION}/site-packages/notebook/static/custom/
+
+# Add endpoint to fetch container uptime
+COPY aiidalab_uptime.py ${CONDA_DIR}/lib/python${PYTHON_MINOR_VERSION}/site-packages/
```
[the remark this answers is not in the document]
edan-bainglass: It doesn't have to live there

stack/lab/countdown/custom.js
edan-bainglass: Correct 👍

stack/lab/countdown/custom.js
danielhollas: Hmm, it looks like you create the `culling-countdown` banner element *at the end* of `insertCountdown` function? So is this if condition just guarding against it being inserted twice?

side note: Maybe using 4 spaces for indentation would help readability here, it's hard to keep up with all the nested functions / callbacks :D

edan-bainglass: > Hmm, it looks like you create the `culling-countdown` banner element _at the end_ of `insertCountdown` function? So is this if condition just guarding against it being inserted twice?

Right. Sorry, I didn't have the code in front of me when I replied.

> side note: Maybe using 4 spaces for indentation would help readability here, it's hard to keep up with all the nested functions / callbacks :D

This was done very quickly. I'll have another look and clean it up where appropriate.

stack/lab/countdown/custom.js
```diff
@@ -0,0 +1,116 @@
+require(["base/js/namespace", "base/js/events"], (Jupyter, events) => {
+  const parseLifetimeToMs = (str) => {
+    // Supports MM:SS, HH:MM:SS, or D-HH:MM:SS formats
+    const regex = /^(?:(\d+)-)?(?:(\d+):)?(\d+):(\d+)$/;
```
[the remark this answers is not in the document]
edan-bainglass: These APIs tend to work best with standard time formats, of which the one I'm working with here is not. Regex gives me more precise control of the format, as well as a clean way to provide defaults. This is true in both languages.

stack/lab/aiidalab_uptime.py
```diff
@@ -0,0 +1,32 @@
...
+            etime = output.decode("utf-8").strip()
+            seconds = self._parse_etime_to_seconds(etime)
+            self.finish({"uptime": seconds})
+        except Exception as e:
+            self.set_status(500)
+            self.finish({"error": str(e)})
+
+    def _parse_etime_to_seconds(self, etime):
+        # Supports MM:SS, HH:MM:SS, or D-HH:MM:SS formats
+        match = re.match(r"(?:(\d+)-)?(?:(\d+):)?(\d+):(\d+)", etime)
```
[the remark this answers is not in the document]
edan-bainglass: Same comment as above

stack/lab/before-notebook.d/70_prepare_countdown_config.sh
edan-bainglass: Hmm, I think I'll keep this as is for this version. Happy to discuss this with the team at the next meeting.

stack/lab/countdown/custom.js
edan-bainglass: See next comment

PR #555, opened 2026-03-24
Misc cleanup and updates

danielhollas:
Some of these are collected from other open PRs so we can keep them clean from unrelated changes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,7 +48,7 @@ jobs:
                   cache-image: true
 
             - name: Set up Docker Buildx
-              uses: docker/setup-buildx-action@v3
+              uses: docker/setup-buildx-action@v4
 
             - name: Login to GitHub Container Registry 🔑
               uses: docker/login-action@v4
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,7 +51,7 @@ jobs:
               uses: docker/setup-buildx-action@v3
 
             - name: Login to GitHub Container Registry 🔑
-              uses: docker/login-action@v3
+              uses: docker/login-action@v4
               with:
                   registry: ghcr.io
                   username: ${{ github.actor }}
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -51,7 +51,7 @@ jobs:
               uses: docker/setup-buildx-action@v3
 
             - name: Login to GitHub Container Registry 🔑
-              uses: docker/login-action@v3
+              uses: docker/login-action@v4
```
danielhollas, line 54: This gets rid of warnings in CI

stack/lab/before-notebook.d/60_prepare-aiidalab.sh
```diff
@@ -6,11 +6,6 @@ set -x
 # Environment.
 export SHELL=/bin/bash
 
-# Fix https://github.com/aiidalab/aiidalab-docker-stack/issues/225
```
danielhollas, line 9: Super old issue which is not relevant in the current image

danielhollas:
There are still some warnings coming from akhilerm/tag-push-action@v2.2.0

https://github.com/aiidalab/aiidalab-docker-stack/actions/runs/23502596467?pr=555

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: akhilerm/tag-push-action@v2.2.0. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

I opened an issue on their repo: https://github.com/akhilerm/tag-push-action/issues/371

Otherwise this PR is ready

yakutovicha (approved):
LGTM!

PR #556, opened 2026-03-25
Cleanup docker-compose config files

danielhollas:
It's pointless to specify DOCKER_STACKS_JUPYTER_CMD envvar since that is controlled by the Dockerfile, and cannot really be changes externally without breaking the image.

Split from #479


yakutovicha (approved):
Tested locally, things work fine 👍. Thanks, @danielhollas!

PR #560, opened 2026-04-08
Pin numpy to 1.x

danielhollas:
The current bokeh version in AWB is incompatible with numpy 2.x There might also be other issues in existing apps so for now we'll pin numpy to 1.x

See https://github.com/aiidalab/aiidalab-widgets-base/issues/733 for more detail


danielhollas:
Notably, we've already pinned numpy in QeApp when testing with aiida-core 2.8

https://github.com/aiidalab/aiidalab-qe/pull/1466

edan-bainglass (approved):
Same comment as [here](https://github.com/aiidalab/aiidalab-widgets-base/pull/740#pullrequestreview-4075823735)

PR #562, opened 2026-04-14
Minor CI cleanup

danielhollas:
Couple things here:

- Fix integration tests after the AWB 3.0.0a0 release
- remove the `runner_aarch64` directory that was used for setting up self-hosted GHA arm runner --- this is no longer needed  :tada: 
- Rename jobs labels: amd64->x86; arm64-> arm. x86/arm are much more visually distinguishable then amd64/arm64!
- minor dedupliction in `main.yml` now that we treat x86 and arm on equal footing

### Before

image

### After

image


edan-bainglass (approved):
LGTM!

PR #576, opened 2026-07-21
quote the bake metadata heredoc

danielhollas:
Same change as was done in aiidateam/aiida-core#7449, see that PR for details.

Notably, the metadata's buildx.build.provenance embeds the push event payload, commit messages included, so every backtick in a commit message is treated as command substitution!


edan-bainglass (approved):
LGTM! @danielhollas how did this not create issues before?

danielhollas:
I was surprised as well 😅

PR #577, opened 2026-07-21
Require Python 3.12 for tests

danielhollas:
Now that the jupyter image runs on Python 3.12, we might as well require minimum Python 3.12 to run the tests to keep things consistent. This also allows us to remove a bunch of transitive dependencies in requirements.txt that were needed for python < 3.11 (e.g. tomli).

I also bumped the pytest dependency to latest.

I regenerated the lockfile with 

```
uv pip compile -U --universal --python 3.12 requirements.in -o requirements.txt
```


yakutovicha (commented):
LGTM!

yakutovicha (approved):

PR #579, opened 2026-07-24
Fix removal of temporary notebooks

### LINKED ISSUE #578: Startup script bug
The [prepare-aiidalab.sh](https://github.com/aiidalab/aiidalab-docker-stack/blob/main/stack/lab/before-notebook.d/60_prepare-aiidalab.sh) startup script contains the following command:

https://github.com/aiidalab/aiidalab-docker-stack/blob/b7919e4ee27ef8c1171e5e437e249855f99a3e71/stack/lab/before-notebook.d/60_prepare-aiidalab.sh#L54-L55

The glob pattern passed to `-name` is unquoted, so the shell expands it before `find` is executed. In case matching files exist in the current working directory, the command becomes invalid (due to the shell command expansion):

```shell
find ... -name .NscfEigenvalues-0.ipynb .NscfEigenvalues-1.ipynb ...
```

This sometimes causes container startup failures (from a recent example):

```
+++ find -L /home/jovyan -maxdepth 3 -name apps_meta.sqlite -writable -delete
+++ find -L /home/jovyan/apps -maxdepth 2 -type f -name .NscfEigenvalues-0.ipynb .NscfEigenvalues-1.ipynb -writable -delete
find: paths must precede expression: `.NscfEigenvalues-1.ipynb'
find: possible unquoted pattern after predicate `-name'?
```

The fix is to simply quote it like so:

```shell

find -L /home/${NB_USER}/apps -maxdepth 2 -type f -name '.*.ipynb' -writable -delete
```

yakutovicha:
fix #578

Update the notebook wildcard so bash doesn't interpret it.

### DIFF
--- a/stack/lab/before-notebook.d/60_prepare-aiidalab.sh
+++ b/stack/lab/before-notebook.d/60_prepare-aiidalab.sh
@@ -52,7 +52,7 @@ fi
 find -L /home/${NB_USER} -maxdepth 3 -name apps_meta.sqlite -writable -delete
 
 # Remove old temporary notebook files.
-find -L /home/${NB_USER}/apps -maxdepth 2 -type f -name .*.ipynb -writable -delete
+find -L /home/${NB_USER}/apps -maxdepth 2 -type f -name '.*.ipynb' -writable -delete
 
 # Uninstall aiidalab from user packages (if present).
 # Would otherwise interfere with the system package.


danielhollas (approved):
Thanks!

edan-bainglass (approved):
LGTM!

PR #582, opened 2026-07-28
Bump appmode to 1.2.0

edan-bainglass:
Includes the [fix](https://github.com/oschuett/appmode/pull/84) for /files/ redirects.


yakutovicha (approved):
go ahead!

danielhollas (approved):
Thanks for figuring out the fix in appmode @edan-bainglass

PR #585, opened 2026-08-04
Update ruff to v0.16 and use its new default rule set

danielhollas:


stack/base/Dockerfile
```diff
@@ -58,10 +58,8 @@ COPY pip.conf "${CONDA_DIR}/pip.conf"
 # Upgrade pip and mamba to latest
 # Install aiida-core and other shared requirements.
 RUN mamba update -y pip && \
-     mamba install --yes \
-     aiida-core.atomic_tools==${AIIDA_VERSION} \
-     mamba-bash-completion \
-     && mamba clean --all -f -y && \
+     mamba install --yes aiida-core.atomic_tools==${AIIDA_VERSION} mamba-bash-completion && \
```
danielhollas, line 61: This is an unrelated change just to bust the Docker cache so that we get the latest version of pip -- 26.2, per discussion with Edan https://github.com/aiidalab/aiidalab-qe/pull/1540#issuecomment-5180014004

edan-bainglass: What do you mean? Can we not just rebuild with --no-cache?

tests/conftest.py
```diff
@@ -71,7 +71,7 @@ def notebook_service(docker_ip, docker_services):
         docker_services.wait_until_responsive(
             timeout=60.0, pause=0.1, check=lambda: is_responsive(url)
         )
-    except Exception as e:
+    except Exception as e:  # noqa: BLE001
```
edan-bainglass, line 74: What was the reason for the base exception?

edan-bainglass (approved):
LGTM! Just a couple of questions 🙂

stack/base/Dockerfile
danielhollas: You can do that locally, but I meant bypassing the GitHub Actions cache

tests/conftest.py
danielhollas: I honestly don't remember

edan-bainglass: Carry on

stack/base/Dockerfile
edan-bainglass: I'm surprised GitHub Actions doesn't provide some mechanism for one-time cache resets. Knowing you, you already checked? 🤔

danielhollas: It does but is annoying in this case since each Docker layer sees to have its own cache entry and it's nkt clear which one I would have to delete.

https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manage-caches#deleting-cache-entries

The alternative would be to simply sait 10 days after which the cache should be cleared automatically.

